PR #4, opened 2025-02-03
GV4-003: add internationalization support with English and Spanish locales

milesibastos:
https://app.holded.com/projects/cards/preview/GV4-003


src/hooks.server.ts
```diff
@@ -1,10 +1,32 @@
...
+  try {
+    locale.set(resolvedLocale);
+  } catch {
+    init({
+      fallbackLocale: config.defaultLocale,
+      initialLocale: resolvedLocale,
+    });
   }
+
+  event.locals.locale = resolvedLocale;
```
subiabre, line 28: `event.locals` interface does not define any properties by default, please add to the Locals interface at `app.d.ts` the `locale: SupportedLocale` property.

src/routes/[[locale]]/+layout.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  if (!urlLocale) {
+    const browserLocale = getBrowserLocale();
+    const locale = browserLocale || config.defaultLocale;
+    return { locale, ...parentData };
+  }
+
+  if (!isValidLocale(urlLocale) || !hasTranslations(urlLocale)) {
+    const redirectUrl = url.pathname.replace(`/${urlLocale}`, "");
+    throw redirect(301, redirectUrl);
```
Comment (subiabre) [inline src/routes/[[locale]]/+layout.ts:18]: Not sure about the 301 Moved Permanently redirect status. We might add the requested localization in the future, but most browsers will forever keep redirecting out of the localization having cached the 301 redirection. I think a 302 status would achieve the redirection just fine without causing issues to users down the line.

There is also bad redirection loop caused by line #17, where the `redirectUrl` resulting for non valid locales is an empty string `""` instead of `"/"` for the root page.

subiabre (changes requested):
Please see requested changes.

Also, there are unidentified bugs somewhere in the `LocaleSwitcher` component:
- When the locale is determined by the browser preferences it does not properly update the selected locale, "English" remains shown as the selected locale despite the code having determined the locale to be used was "es".
- When trying to switch to "English" from any other locale the URL is not updated for "en", neither is the locale applied for display.

Finally, could you move the component to `src/lib/components` and add basic component story?

Thank you.

PR #17, opened 2025-02-25
GV4-031: Mechanism for saving items in cart to web storage API

milesibastos:


subiabre (approved):
Nice work. Just won't merge it yet into GV4-020

PR #26, opened 2025-03-07
GV4-052: Cuenta atrás de proyecto

milesibastos:


src/lib/components/CountdownLabel/CountdownLabel.svelte
```diff
@@ -0,0 +1,68 @@
...
+            formattedTime = $_("countdown.ended") || "Campaña finalizada";
+            return;
+        }
+
+        // Calculate days, hours, minutes
+        days = Math.floor(diff / (1000 * 60 * 60 * 24));
+        hours = Math.floor((diff % (1000 * 60 * 60 * 24)) / (1000 * 60 * 60));
+        minutes = Math.floor((diff % (1000 * 60 * 60)) / (1000 * 60));
+
+        // Use Intl.RelativeTimeFormat for human-readable relative time
```
subiabre, line 30: Current usage of Intl rtf is not compliant with the spec. Intl's relative time format adds localization out of our control and the current usage only takes care of the largest unit. Please, make the component always display:

```
XXd XXh XXm
```

src/routes/[[locale]]/projects/[projectId]/+server.ts
```diff
@@ -119,14 +119,18 @@ const map = (
         },
     };
 
-    const { id, title, subtitle, description, territory, category } = project;
+    const { id, title, subtitle, description, territory, category, status, calendar } = project;
     const territoryLabel = territory.subLvl2 || territory.subLvl1 || territory.country || "";
 
+    const deadline = calendar && project.deadline ? calendar[project.deadline] : null;
```
Comment (subiabre) [inline src/routes/[[locale]]/projects/[projectId]/+server.ts:125]: When the deadline is of case `optimum`, the countdown should still go for the minimum deadline. This deadline is mandatory and failure to raise minimum funds inside the minimum deadline will terminate the project regardless of the optimum deadline.

The optimum deadline should only be watched when a Project with case `optimum` deadline is still in campaign after the minimum deadline was surpassed.

subiabre (changes requested):

subiabre (approved):

PR #29, opened 2025-04-01
GV4-076: Implement Checkout Process

jpiere-vailos:


src/components/RegisterC.astro
subiabre, line 1: Why the `C` suffix at the end of these filenames?

src/stores/cart.ts
```diff
@@ -0,0 +1,134 @@
+import { writable, derived } from "svelte/store";
+
+export interface CartItem {
+    key: string;
+    title: string;
+    amount: number;
+    quantity: number;
+    image?: string;
+    project?: number;
+    owner?: string;
```
subiabre, line 10: This property is very confusing. As this interface represents a cart item I would assume the owner of an item in a cart is the owner of such cart, i.e: the user. This is not the case here because what is being stored in here is the API resource to which this cart item will _target_ to in the charge item in the API checkout. Please rename accordingly.

src/stores/cart.ts
```diff
@@ -0,0 +1,134 @@
...
+}
+
+interface CartStore {
+    items: CartItem[];
+}
+
+function generateKey(item: Omit<CartItem, "key">): string {
+    const base = `${item.title}-${item.project ?? "none"}-${item.owner ?? "none"}`;
```
subiabre, line 18: Because the owner of the item is related to the target resource, doing `item.project`+`item.owner` is basically redundant, the `owner` in an item to `project` _X_ will always be _Y_.

And since cart items can go to projects or to a tipjar I would remove the project part and just leave the owner part.

Moreso, this key gen strategy feels vulnerable to collisions, so adding a third element for uniqueness would be great. A date timestamp of item addition to the cart should be enough.

Finally, I would change the order of properties altogether from larger to smaller:

`item.owner` + `item.title` + `item.date`.

This allows for grouping items by project (owner) by just doing an alphabetic sort.

src/stores/cart.ts
```diff
@@ -0,0 +1,134 @@
...
+    image: "",
+    owner: "Platoniq",
+};
+
+const defaultItem = { ...defaultDonation, key: generateKey(defaultDonation) };
+
+const isBrowser = typeof window !== "undefined";
+
+function loadInitialCart(): CartStore {
+    if (!isBrowser) return { items: [defaultItem] };
```
subiabre, line 35: I'm not comfortable with hardening the default tip item this way. Leaving the checkbox marked by default in the markup should be enough. Otherwise we are making it harder to change the "defaultness" and we could be introducing bugs that add the tip even after users have opted out of tip via checkbox.

src/utils/extractId.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Extracts the ID from an IRI string like "/v4/users/142" or "https://api.goteo.org/v4/accountings/2533".
+ *
+ * @param iri - A string representing an IRI (Internationalized Resource Identifier)
+ * @returns The extracted ID as a string, or null if not available
+ */
+export function extractId(iri?: string): string | null {
+    if (!iri) return null;
+    const parts = iri.split("/").filter(Boolean);
+    return parts.length ? parts[parts.length - 1] : null;
```
subiabre, line 10: Smart work! However it should consider cases where the ID is not the last piece, like when obtaining person or organization data from a user: /v4/users/123/person

subiabre (changes requested):

PR #31, opened 2025-04-15
GV4-087: Create checkout with local data

jpiere-vailos:


src/components/Checkout/FeedbackForm.svelte
```diff
@@ -0,0 +1,113 @@
+<script lang="ts">
+    import { onMount } from "svelte";
+    import { t } from "../../i18n/store";
+
+    onMount(() => {
+        const select = document.getElementById("campaign") as HTMLSelectElement | null;
+        const textarea = document.getElementById("review-message") as HTMLTextAreaElement | null;
```
subiabre, line 7: Why not use the `bind` directive? https://svelte.dev/docs/svelte/bind

src/components/Checkout/Summary.svelte
```diff
@@ -17,16 +17,24 @@
...
-        de aporte a la fundación
-    </p>
-    <p>{$t("checkout.summary.deduct")}</p>
+<div class="flex flex-col gap-6 px-6 pt-6 pb-0">
+    <div>
+        <h2 class="text-[32px] font-semibold text-[#462949]">
+            {$t("checkout.summary.total.title")}
+        </h2>
+        <p class="text-[56px] leading-tight font-bold text-[#462949]">
+            {formatCurrency($total, "EUR", { showSymbol: true })}
```
subiabre, line 26: Can we make "EUR" an environment var instead?

subiabre (changes requested):

PR #32, opened 2025-04-23
GV4-091: Handle return from payment gateway

jpiere-vailos:


src/components/Checkout/Tipjar.svelte
```diff
@@ -1,63 +1,126 @@
...
     let initialized = false;
 
-    const createDonationItem = (amount: number) => ({
+    const createDonationItem = (amount: number): Omit<CartItem, "key"> => ({
         title: "Donación Platoniq",
-        amount: Math.round(amount * getUnit("EUR")),
+        amount: Math.round(amount * getUnit(defaultCurrency)),
         quantity: 1,
         image: "",
         target: "Platoniq",
```
subiabre, line 23: Having a property `target` in a data schema that in turns will be used for Checkout with the API lends itself to confusion when our `target` here is a string literal and the `target` the API expects is a IRI reference or accounting ID. We should store `accountingIdPlatoniq` here.

src/components/Checkout/Tipjar.svelte
```diff
@@ -1,63 +1,126 @@
...
         if (!initialized) {
             const item = $cart.items.find((i) => i.target === "Platoniq");
             if (item) {
-                value = item.amount / getUnit("EUR");
-                previousValue = value;
+                value = item.amount / getUnit(defaultCurrency);
+                isChecked = true;
+            } else {
+                value = 3;
+                isChecked = false;
```
subiabre, line 36: Should be checked by default.

subiabre (changes requested):

src/components/Checkout/Cart.svelte
```diff
@@ -3,30 +3,23 @@
...
-        const newPath = `/${currentLang}/payment`;
-        window.location.href = newPath;
-    }
-
     const groupedByOwner = derived(items, ($items) => {
         const grouped = $items.reduce(
             (acc, item) => {
                 const key = item.target || "";
-                if (key.toLowerCase().trim() === "platoniq") return acc;
+                if (key.toLowerCase().trim() === accountingIdPlatoniq) return acc;
```
subiabre, line 22: Please notice that our key is now a true key from the API, mutating it breaks our reliability on these keys to remain unique.

src/components/Checkout/Cart.svelte
```diff
@@ -38,6 +31,37 @@
...
+        }
+        return "";
+    }
+
+    async function loadDisplayNames() {
+        const names: Record<string, string> = {};
+        const $grouped = get(groupedByOwner);
+
+        for (const [target, items] of Object.entries($grouped)) {
+            if (target.toLowerCase().trim() === "platoniq") continue;
```
subiabre, line 55: This if statement will never be true because we are now storing accounting IDs as keys, which are integers. We need to check `target === accountingIdPlatoniq`.

src/stores/cart.ts
```diff
@@ -9,7 +9,6 @@ export type CartItem = {
     project?: number;
     target: string;
```
subiabre, line 10: We are storing accounting IDs here, which are numbers.

subiabre (changes requested):

subiabre (approved):

PR #34, opened 2025-04-24
GV4-094: Fixes and improvements in the single project upper section

jpiere-vailos:


src/i18n/utils.ts
```diff
@@ -20,14 +20,47 @@ function getNestedValue(obj: Record<string, unknown>, path: string): string | un
...
+        vars?: Record<string, string | number>,
+        allowHTML: boolean = false,
+    ): string => {
+        let text =
+            getNestedValue(labels[lang], key) ?? getNestedValue(labels[defaultLang], key) ?? key;
+
+        if (vars) {
+            Object.entries(vars).forEach(([varKey, value]) => {
+                const safeValue = allowHTML ? String(value) : escapeHTML(String(value));
+                text = text.replace(new RegExp(`{${varKey}}`, "g"), safeValue);
```
subiabre, line 41: Please use double braces (`{{` and `}}`) to open and close interpolated strings. Braces is still too common to rely on a single instance of it as a string that should be interpolated.

src/components/CartButton.svelte
```diff
@@ -0,0 +1,42 @@
...
+        });
+        return unsubscribe;
+    });
+</script>
+
+<button
+    type="button"
+    on:click={handleDirect}
+    class="group relative inline-flex cursor-pointer items-center"
+    aria-label="Ir al checkout"
```
subiabre, line 31: User-facing text, should be localized.

subiabre (changes requested):

src/i18n/locales/es.json
```diff
@@ -44,13 +47,19 @@
...
-      "description": "Goteo pertenece a la Fundación Platoniq. Al colaborar en una campaña tu dinero será custodiado por Platoniq y donado a Climática en caso de llegar a los objetivos de éxito marcados.",
-      "learnMore": "Saber más."
-    },
     "impact": {
       "description": "Al colaborar con un proyecto impulsado en Goteo puedes solicitar un certificado de donación. Así, te deduces hasta un 80% de lo que hayas aportado en tu próxima declaración de la renta.",
       "learnMore": "Saber más."
+    },
+    "banner": {
+      "donator": {
+        "description": "Goteo pertenece a la Fundación Platoniq. Al colaborar en una campaña tu dinero será custodiado por Platoniq y donado a {owner} en caso de llegar a los objetivos de éxito marcados.",
```
subiabre, line 56: Remember we are using double braces now.

subiabre (changes requested):

PR #37, opened 2025-04-29
GV4-085: Enable routing without locale prefix for default language, detect preferred language from Accept-Language

jpiere-vailos:


src/middleware/utils.ts
```diff
@@ -0,0 +1,115 @@
+import type { APIContext } from "astro";
+
+const defaultLang = import.meta.env.PUBLIC_LANGUAGE_DEFAULT || "es";
```
subiabre, line 3: Plase add the variable declaration along with default "es" value in the .env file instead of code.

src/middleware/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * Builds a clean redirect URL by combining the language code and pathname.
+ * Ensures no double slashes.
+ *
+ * @param {string} lang - The selected language (e.g., "en", "es", "ca").
+ * @param {string} pathname - The current request pathname.
+ * @returns {string} - A safe redirect URL.
+ */
+export function buildRedirectUrl(lang: string, pathname: string): string {
+    const cleanPath = pathname === "/" ? "" : pathname;
```
subiabre, line 14: A much safer strategy here is to strip all leading slashes, something like `replace(/^\/+/, '')`. Then we can add the single leading slash safely to build paths like `/${lang}/${path}`. If trailing slashes are a problem we apply the same strategy and replace all trailing slashes with a regex like `/\/+$/`.

```js
path.replace(/^\/+/, ''); // 'path', '/path' and '//path' -> 'path'
path = `/${lang}/${path}`; // '/es/path';

return path.replace(/\/+$/, ''); // '/es/' -> '/es'
```

This allows to pass a path argument to the function without having to care if our path has bad leading or trailing slashes.

src/middleware/index.ts
```diff
@@ -1,55 +1,37 @@
...
 import { languagesList } from "../i18n/locales/index";
 import { useTranslations } from "../i18n/utils";
 
 import type { Locale } from "../i18n/locales/index";
 import type { APIContext } from "astro";
 
-export const onRequest = defineMiddleware((context: APIContext, next) => {
-    const url = new URL(context.request.url);
-    const pathParts = url.pathname.split("/").filter(Boolean);
+export const onRequest = defineMiddleware(async (context: APIContext, next) => {
```
subiabre, line 10: Bad indentation! Be sure to run prettier on your PRs

src/middleware/utils.ts
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Detects the appropriate locale based on URL, cookie or Accept-Language header.
+ * Redirects if needed.
+ */
+export function detectLocale(
+    context: APIContext,
+    maybeLang: string | undefined,
+    validLangs: string[],
+): string {
```
subiabre, line 27: This function signature is heavily tied to usage in index.ts middleware. Please update arguments for better re-usability of function. `maybeLang` is actually first path segment, which we can obtain from the context argument so it can be deleted.

Then having code in the function body to get the `preferredLanguage` from the context URL standardizes the locale obtention flow for other function calls anywhere in the code with access to the context:

1. User-defined request path. If not present, then...
2. Browser-defined cookie. If not present, then...
3. Browser defined request header. If not present, then...
4. Application default.

subiabre (changes requested):

subiabre (changes requested):

subiabre (approved):

PR #42, opened 2025-05-21
GV4-113: Rename route and support slug project

jpiere-vailos:


src/pages/[...lang]/projects/[idOrSlug].astro
Comment (subiabre) [inline src/pages/[...lang]/projects/[idOrSlug].astro:1]: Current route pattern is `project/<slug>`. Please rename the directory to `project` in order to update the route.

subiabre (changes requested):

subiabre (approved):

PR #46, opened 2025-05-28
GV4-112: Implement filtering in contributions viewer

jpiere-vailos:


src/i18n/locales/es.json
```diff
@@ -413,6 +413,61 @@
       "search": {
         "placeholder": "Buscar...",
         "btnLabel": "Buscar"
+      },
+      "paymentMethod": {
+        "title": "Método de pago",
+        "options": {
+          "all": "Todos",
+          "paypal": "Paypal",
+          "stripe": "Tarjeta",
```
subiabre, line 422: Stripe's name is `Stripe`. `Tarjeta` is a placeholder name we use for direct-to-bank gateways like Redsys, which is not yet available in the v4 API.

subiabre (changes requested):

PR #57, opened 2025-06-17
GV4-135: Campaign budget slider in tab

jeanpieredev:


src/components/Carousel.svelte
```diff
@@ -0,0 +1,213 @@
...
+    export let dimNext: boolean = false;
+
+    let container: HTMLDivElement;
+    let totalItems = 0;
+    let activeGroupValue = 0;
+    const activeGroup = writable(0);
+
+    let totalGroups = 0;
+    let isAtStart = true;
+    let isAtEnd = false;
```
subiabre, line 19: This property is not being properly updated. The right scroll arrow needs to be shown disabled when there are no more items to the right of the container scrollLeft, which is currently failing on first loads

subiabre (changes requested):
Please fix `isAtEnd` computing and merge conflicts.

PR #81, opened 2025-07-31
GV4-159: Implement user module test cases for admin panel and project progress

sdelafuente2013:


cypress/e2e/admin_view_individual_contributions_breakdown.cy.ts
```diff
@@ -0,0 +1,106 @@
...
+            body: [
+                { name: "stripe", title: "Stripe" },
+                { name: "paypal", title: "PayPal" },
+            ],
+        }).as("gatewaysMock");
+
+        cy.intercept("GET", "**/v4/gateway_charges**", {
+            statusCode: 200,
+            body: {
+                member: [
```
subiabre, line 16: The schema of these items does not match the schema of the Charge resources from the API

subiabre (changes requested):

PR #94, opened 2025-08-28
Update component heights to `h-full`

sdelafuente2013:


src/components/project/SingleProject.svelte
```diff
@@ -93,7 +93,7 @@
         </div>
     </div>
 
-    <div class="mb-6 flex items-stretch justify-between gap-6">
+    <div class="mb-6 flex h-150 items-stretch justify-between gap-6">
```
subiabre, line 96: Please, try to use `h-132` as it seems a better fit for 16:9 videos.

subiabre (changes requested):

PR #124, opened 2025-09-29
feat: search page

milesibastos:
image

This pull request migrates the campaign card component from Astro to Svelte, updates its usage across the codebase, and introduces several new reusable UI components for the search interface. The main themes are the migration to Svelte for improved interactivity and consistency, and the addition of modular search/filter UI elements.

**Migration to Svelte components:**

* Converted the campaign card from Astro to Svelte, preserving its functionality and responsive design. (`CampaignCard.astro` removed, `CampaignCard.svelte` added) [[1]](diffhunk://#diff-3d6d268d2e646521982d0ee49b43337c2d1d0211d27a946e25472c07cc9c5392L1-L126) [[2]](diffhunk://#diff-35dcb8e06e6dc0839a22ba1bae3a575048e581a4fe0ab0cc32fadce9f26d4369R1-R102)
* Updated all usages of `CampaignCard.astro` to use the new Svelte version in `Campaigns.astro` and `CampaignsTabs.astro`. [[1]](diffhunk://#diff-a1d3fc0f2d3d532ff6ddd878dc8eae72f8047a669a43cb4a908c9483a03034eaL2-R2) [[2]](diffhunk://#diff-ad938a24754c70e0027337abc19aaafb7f010130b6f083b8c0e66f701be668d1L2-R2)

**New search/filter UI components:**

* Added `CategoryFilter.svelte`, a pill-style interactive category selector matching Figma designs, with accessibility and SSR hydration support.
* Added `FilterDropdown.svelte`, a reusable dropdown for filter options (period, status, location), featuring accessibility, keyboard navigation, and collision-free IDs.
* Added `SearchButton.svelte`, a customizable button supporting primary, secondary, and ghost variants, with loading and sizing options.
* Added `LoadingSpinner.svelte`, a reusable loading indicator component with ARIA attributes for accessibility.

**UI consistency improvements:**

* Refactored button class order in `HeaderButtons.svelte` for consistency with design system.

### DIFF
--- a/src/components/profile/ProfileProjectsList.astro
+++ b/src/components/profile/ProfileProjectsList.astro
@@ -1,6 +1,6 @@
 ---
 import type { Campaign } from "../../types/campaign";
-import CampaignCard from "../home/CampaignCard.astro";
+import CampaignCard from "../home/CampaignCard.svelte";
 
 export interface Props {
     campaigns: Campaign[];


src/pages/[...lang]/search.astro
```diff
@@ -0,0 +1,68 @@
...
+import type { Project } from "../../openapi/client/types.gen";
+
+const { t, lang } = Astro.locals;
+
+// Get initial projects from API (for SSR)
+let initialProjects: Project[] = [];
+
+try {
+    // Fetch initial projects from API
+    initialProjects = await projectsService.getInitialProjects(lang, {
```
Comment (subiabre) [inline src/pages/[...lang]/search.astro:16]: This call here can delay the initial page load, I feel like we could populate the inititalProjects after the client has finished the initial page load by instead triggering a default search query.

src/components/search/SearchResultsClient.svelte
```diff
@@ -0,0 +1,278 @@
...
+                {$t("search.results.found", { count: $resultCount })}
+            </p>
+        </div>
+    {:else if $hasInitialResults}
+        <div class="mb-6">
+            <p class="text-secondary text-lg" data-testid="initial-result-count">
+                {$t("search.showing_top_results", { count: $resultCount })}
+            </p>
+        </div>
+    {/if}
```
subiabre, line 185: This block is very problematic. The result count is not being properly updated and neither are we instructed to place a clear all filters actionable here. The filter clearing is in fact deliberately left inside the search card, which should stick to the screen just as, and below, the sticky header.

src/components/search/SearchResultsClient.svelte
```diff
@@ -0,0 +1,278 @@
...
+                    </div>
+                {/each}
+            </div>
+        {/if}
+    </div>
+
+    <!-- Pagination Controls -->
+    {#if $hasActualSearchResults && !$isSearching}
+        <div class="mt-8">
+            <SearchPagination showSummary={true} class="search-pagination" />
```
subiabre, line 217: I'm very sorry but as fine as the pagination in this component is, we are again not instructed in the Figma design to have such fine control over the pagination. The available behaviour instead is a "load more" button, which should append on this results client component the next page of results until the last page.

subiabre: We could still find a use for the detailed pagination component, so please leave that one as it is. Thanks.

src/services/projectsService.ts
```diff
@@ -0,0 +1,118 @@
...
+    async getProjectsBySlugs(
+        slugs: string[],
+        options?: {
+            abortSignal?: AbortSignal;
+        },
+    ): Promise<Project[]> {
+        try {
+            const response = await apiProjectsGetCollection({
+                query: {
+                    "slug[]": slugs,
```
subiabre, line 97: If you want to get a single Project, you need to call `apiProjectsIdGet`, which takes an id or slug as path identifier, this call will in fact retrieve a collection of projects. The `slug` filter is meant to allow for text searches of project slugs just like you would search for titles and descriptions.

src/utils/projectTransform.ts
```diff
@@ -0,0 +1,37 @@
...
+    const projectAny = project as any;
+
+    // Get image URL - simple fallback to video thumbnail
+    const image = projectAny.cover || project.video?.thumbnail || "";
+
+    // Get minimum budget
+    const minimum: ApiMoney = project.budget?.minimum?.money || { amount: 0, currency: "EUR" };
+
+    // Get raised amount from accounting
+    const obtained: ApiMoney = projectAny.accounting?.balance || { amount: 0, currency: "EUR" };
```
subiabre, line 20: Will not work. The `accounting` here is a related resource, in this case the API does not retrieve the full resource but instead it gives us the IRI of the resource, that means `project.accounting` is a string, which you would see in the IDE if this wasn't a proxy object of type `any`. We need to retrieve the actual data of that accounting from the API.

You can do so by extracting the accounting id from the IRI string and then fetching the accounting. Example below should get you pretty close to that.

```js
const { data: accounting } = await apiAccountingsIdGet({ path: { id: extractId(project.accounting) });
```

subiabre (changes requested):

src/components/search/SearchResultsClient.svelte
subiabre: For reference here is the Figma frame
image

PR #205, opened 2026-02-06
[feat] RisksCard component

Dayv1dDev:
I have to point out that some i18n keys are incomplete due to I don't know if they are created atm, so for now I'm leaving it like this. Additionally, there are some TODO comments inside the component that reference the functionalities we would have to add in the future for this component. Right now, I focused this task in styling + add possible props this component could handle by his own.


src/i18n/locales/es.json
```diff
@@ -1279,6 +1279,19 @@
       }
     }
   },
+  "reviews": {
```
subiabre: Please, adapt these keys to the new standard for translation keys.

1. `reviews.risks-cards.tags` to `domain.review.risks`
2. `reviews.risks-cards.btns` to `pages.review.XX`

src/components/library/RisksCard.svelte
```diff
@@ -0,0 +1,57 @@
...
+    import type { ProjectReviewArea } from "../../openapi/client";
+
+    let { review, newMessage = false } = $props<{
+        review: ProjectReviewArea;
+        newMessage?: boolean;
+    }>();
+
+    let newMsgCardShadow =
+        "0 35px 10px 0 rgba(0, 0, 0, 0.00), 0 22px 9px 0 rgba(0, 0, 0, 0.01), 0 13px 8px 0 rgba(0, 0, 0, 0.05), 0 6px 6px 0 rgba(0, 0, 0, 0.09), 0 1px 3px 0 rgba(0, 0, 0, 0.10)";
+    let tagVariant: any = $state();
```
subiabre: Please, implement the variants of this component like the rest of the components do. Use an object with keys for the styles of each variant and receive the variant via prop typed as the keys of said object. You can make this prop undefined by default and decide the default variant internally from the review risk.

subiabre (changes requested):

subiabre (approved):

PR #206, opened 2026-02-06
[fix] Review FeedbackForm and PaymentMethod components

Dayv1dDev:
.


subiabre:
The codebase of this PR was far too outdated even after attempting to merge `develop` in its latest state was too problematic. We will restart the progress of this PR in new tasks.

PR #208, opened 2026-02-13
[feat] Basic & multiselect dropdown menu component with search optionality

Dayv1dDev:
In this PR I implemented the Dropdown Menu component, which has basic and multiselect variants + optionality to add search if needed with only adding `hasSearch` in the component call.

**Pending:** add Checkbox component which is implemented in another PR + add functionality and logic to multiselect, searching and basic dropdown menu buttons.


src/components/library/Dropdown/dropdown.types.ts
```diff
@@ -0,0 +1,8 @@
+export type DropdownVariant = "basic" | "multiselect";
+
+export type DropdownItemPosition = "start" | "middle" | "end";
```
subiabre: I think this entire type can be discarded by allowing the `DropdownItem` component to take in class names as a prop and then passing the appropriate `rounded-*` class in `DropdownMenu`.

Not only is having components with a `class` prop which allows to override the default styles via `twMerge` already a common practice across the rest of the components and a feature missing in this component, I also think it would result in a simpler solution.

src/components/library/Dropdown/DropdownMenu.svelte
```diff
@@ -0,0 +1,52 @@
...
+
+    function getPosition(index: number, length: number): DropdownItemPosition {
+        if (index === 0 && !hasSearch) return "start";
+        if (index === length - 1) return "end";
+        return "middle";
+    }
+</script>
+
+<div
+    class="flex w-full max-w-[416px] flex-col self-center rounded-lg bg-transparent shadow-[0_35px_10px_0_rgba(0,0,0,0),0_22px_9px_0_rgba(0,0,0,0.01),0_13px_8px_0_rgba(0,0,0,0.05),0_6px_6px_0_rgba(0,0,0,0.09),0_1px_3px_0_rgba(0,0,0,0.1)]"
```
subiabre: Why is there a max width? Maybe there is a problem I am not seeing by allowing dropdowns to take as much horizontal space as it's parent allow it. But why is it hard-coded to 416px? At least use a relative size like `1xl` to allow for appropriate scaling across screens.

subiabre (changes requested):

src/components/library/Dropdown/dropdown.types.ts
Dayv1dDev: Even with that I think the type can't be discarded because will be used anyway in both `DropdownMenu` and `DropdownItem` components (for the conditional rendering). And I didn't want to use `twMerge` because it consumes resources that I thought (in this case) were unnecessary. Tell me if I am/was wrong.

subiabre (approved):

PR #218, opened 2026-03-17
Feat: implement component PasswordInput

pepvillalonga:


src/components/library/PasswordInput.svelte
```diff
@@ -0,0 +1,61 @@
+<script lang="ts">
+    import Eye from "../../svgs/Eye.svelte";
+
+    interface Props {
+        class?: string;
```
subiabre, line 5: If you use the type `ClassNameValue` from `tailwind-merge` it will be taken not just as an string, but as CSS class names and it will allow the extensions in your IDE to know tailwind classes are written in here.

src/components/library/PasswordInput.svelte
```diff
@@ -0,0 +1,61 @@
...
+        placeholder?: string;
+        name?: string;
+        id?: string;
+        label?: string;
+    }
+
+    let {
+        class: classes = "",
+        value = $bindable(""),
+        placeholder = "Escribe tu contraseña",
```
subiabre, line 16: Do not hardcode values. Please, use the `t` function from `src/i18n/store`. Keep in mind that, since this function is given to our svelte code as an [store](https://svelte.dev/docs/svelte/stores), it needs to be called with the dollar sign prefix (`$t`).

src/components/library/PasswordInput.svelte
```diff
@@ -0,0 +1,61 @@
...
+
+    let showPassword = $state(false);
+
+    function togglePassword() {
+        showPassword = !showPassword;
+    }
+</script>
+
+<div
+    class="border-secondary relative flex h-14 w-full max-w-[432px] items-center justify-between rounded-3xl border bg-white p-4 {classes}"
```
subiabre, line 30: The basic append of the CSS classes given in the `classes` prop will not actually work as expected. You need to use the `twMerge` function from `tailwind-merge` which will actually resolve precedence and override conflicting rules based on order, simply pass the `classes` prop as the last argument to allow to override the default classes of the component.

Also: why the `max-w-[432px]`? If the component must grow larger than that somewhere I think it's perfectly reasonable to allow it do so.

src/components/library/PasswordInput.svelte
```diff
@@ -0,0 +1,61 @@
...
+    {#if showPassword}
+        <input
+            type="text"
+            {name}
+            {id}
+            {placeholder}
+            bind:value
+            class="flex-1 border-none bg-white text-black outline-none focus:ring-0"
+        />
+    {:else}
```
subiabre, line 47: Most of the markup is the same for both cases. It could be resolved with much less code by simply switching the values of the concerned attributes for each case with ternary operators, take the following example:

```svelte
<input type={showPassword ? "text" : "password"}>
```

src/svgs/Eye.svelte
```diff
@@ -0,0 +1,55 @@
...
+            d="M10.5 8C10.5 9.38071 9.38073 10.5 8.00002 10.5C6.61931 10.5 5.50002 9.38071 5.50002 8C5.50002 6.61929 6.61931 5.5 8.00002 5.5C9.38073 5.5 10.5 6.61929 10.5 8Z"
+        />
+        <path
+            fill-rule="evenodd"
+            clip-rule="evenodd"
+            d="M1.58399 3.72265C1.73717 3.49289 2.0476 3.4308 2.27737 3.58397L14.2774 11.584C14.5071 11.7372 14.5692 12.0476 14.416 12.2774C14.2629 12.5071 13.9524 12.5692 13.7227 12.416L1.72267 4.41603C1.4929 4.26285 1.43082 3.95241 1.58399 3.72265Z"
+        />
+    </svg>
+{:else}
+    <svg
```
subiabre, line 38: Only the `<path>` tags change in each case, the `<svg>` tag and its attributes remain the same for both `closed` true or false. The if clauses could be written inside the generic `<svg>` tag.

subiabre (changes requested):

PR #221, opened 2026-03-19
Feat: implement component Search

pepvillalonga:


src/components/library/Search.svelte
```diff
@@ -0,0 +1,52 @@
+<script lang="ts">
+    import SearchIcon from "../../svgs/SearchIcon.svelte";
+
+    interface Props {
+        class?: string;
```
subiabre: Remember to use the better type `ClassNameValues` from `tailwind-merge`.

src/components/library/Search.svelte
```diff
@@ -0,0 +1,52 @@
...
+        placeholder = "Buscar...",
+        name = "search",
+        id = "search",
+        label = undefined,
+        onsubmit,
+    }: Props = $props();
+</script>
+
+<div
+    class="border-secondary relative flex h-14 w-full max-w-[432px] items-center justify-between rounded-3xl border bg-white p-4 {classes}"
```
subiabre: Again, remember to use the function `twMerge` from `tailwind-merge` to merge classes which are expected to have any kind of complexity on them. This kind of simple class definition from outside is more appropriate for simple components where reasonable use case only involves changing the size of the default component.

Also, why the `max-w` property and if the component is better with a default max width, use relative units. I think if you have the [Tailwind CSS IntelliSense](https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss) extension for VSCode it will already suggest you the nearest relative unit definition.

subiabre (changes requested):

subiabre (approved):

PR #222, opened 2026-03-20
Feat: implement component validation Email

pepvillalonga:


src/components/library/Email.svelte
```diff
@@ -0,0 +1,78 @@
...
+    const emailRegex = /^[^@].*@.*\.[a-zA-Z]+$/;
+
+    const isValid = $derived(value === "" || emailRegex.test(value));
+</script>
+
+<div class={twMerge("flex w-full flex-col", classes)}>
+    <div
+        class={twMerge(
+            "relative flex h-14 w-full items-center justify-between rounded-3xl border bg-white p-4 transition-all",
+            "border-secondary",
```
subiabre: Why is this class given as a separate argument to `twMerge` ? If it's part of the default set of classes simply give it in the first argument.

src/svgs/CloseIcon.svelte
```diff
@@ -13,14 +13,13 @@
...
-    ><g id="SVGRepo_bgCarrier" stroke-width="0"></g>
-    <g id="SVGRepo_tracerCarrier" stroke-linecap="round" stroke-linejoin="round"> </g>
+>
     <g id="SVGRepo_iconCarrier">
         <g id="Menu / Close_LG">
             <path
                 id="Vector"
                 d="M21 21L12 12M12 12L3 3M12 12L21.0001 3M12 12L3 21.0001"
-                stroke="#462949"
+                stroke="currentColor"
```
subiabre, line 22: Very nice 👍🏼

src/components/library/Email.svelte
```diff
@@ -0,0 +1,78 @@
...
+        placeholder = $t("login.form.email"),
+        name = "email",
+        id = "email",
+        label = undefined,
+        errorMessage = $t("login.error.invalidEmail"),
+    }: Props = $props();
+
+    const emailRegex = /^[^@].*@.*\.[a-zA-Z]+$/;
+
+    const isValid = $derived(value === "" || emailRegex.test(value));
```
subiabre: If the component allowed users to disable the default validation of the component and put in their own it would be nicer. I would like to be able to do something like this:

```svelte
// Parent.svelte
<script lang="ts">
let value: string = "";
let isValid: boolean = false;

$effect(() => {
    isValid = value.endsWith("gmail.com");
});
</script>

<Email bind:value bind:isValid validate={false} errorMessage="Please enter a Gmail address"/>
```

You would need to check if the `validate` prop is true before internally applying any state to the `isValid` prop.

subiabre (changes requested):

subiabre (approved):

PR #224, opened 2026-03-24
Feat: implement static content page

pepvillalonga:


src/components/library/ReturnButton.svelte
subiabre, line 1: Either `ButtonReturn` or `ReturnButton` would be much better component name. "Return" itself is very vague and it overlaps with the very used language keyword "return".

src/components/library/ReturnButton.svelte
```diff
@@ -0,0 +1,31 @@
+<script lang="ts">
+    import { twMerge, type ClassNameValue } from "tailwind-merge";
+
+    import { t } from "../../i18n/store";
+    import Back from "../../svgs/Back.svelte";
+
+    interface Props {
+        class?: ClassNameValue;
+
+        text?: string;
```
subiabre, line 9: Why the extra line between `class` and `text` ?

tsconfig.json
```diff
@@ -2,5 +2,7 @@
   "extends": "astro/tsconfigs/strict",
   "include": [".astro/types.d.ts", "src/types", "**/*"],
   "exclude": ["dist"],
-  "compilerOptions": {}
+  "compilerOptions": {
+    "types": ["astro/client", "svelte"]
```
subiabre: Might break the transpilation step by limiting how much types the TypeScript compiler auto includes for us. See what https://www.typescriptlang.org/tsconfig/#types says about this:

> If types is specified, only packages listed will be included in the global scope.

If this was due to a problem with your IDE understanding the code you were working with you might want to look into installing the Astro extension for VSCode: https://marketplace.visualstudio.com/items?itemName=astro-build.astro-vscode

If you need further configuration, you might want to add a `.vscode/settings.json` file to the repo, this specific config might help you:

```json
{
  "typescript.enablePromptUseWorkspaceTsdk": true
}
```

subiabre (changes requested):

subiabre (approved):

PR #225, opened 2026-03-25
feat: Add TextArea.svelte component

angeldipinto:


src/components/library/TextArea.svelte
```diff
@@ -0,0 +1,65 @@
+<script lang="ts">
+    import { twMerge, type ClassNameValue } from "tailwind-merge";
+
+    interface Props {
+        class?: ClassNameValue;
+        placeholder?: string;
+        labelText?: string;
+        value?: string;
+        helperText?: string;
```
subiabre: Other components often use lean names for props (as short as possible while still being descriptive), also always trying to make them as similar as standard attribute names of HTML tags.

Names like `labelText` and `helperText` break the convention of other components, `label` and `helper` are shorter, still descriptive enough and are very similar if not identical to attributes of some HTML tags.

src/components/library/TextArea.svelte
```diff
@@ -0,0 +1,65 @@
...
+
+    let variantStyles = {
+        default: "border-secondary focus:ring-secondary bg-white",
+        disabled: "border-transparent bg-gray-light cursor-not-allowed",
+        error: "border-tertiary focus:ring-tertiary bg-white text-tertiary focus:ring-1",
+        success: "border-green-500 focus:ring-green-500 bg-white text-green-500",
+    };
+
+    let {
+        variant = "default",
```
subiabre: There is a much better way to make the `default` variant be the default one shown in our component.

Instead of defining it in the `variantStyles` object and setting it's name as the default value of the `variant` prop we can simply just include the styles of the variant in the base arguments of the `twMerge` function call.

This not only achieves our goal in less lines of code, it also makes the component perform better by not forcing `twMerge` to compute the final classes every time.

src/components/library/TextArea.svelte
```diff
@@ -0,0 +1,65 @@
...
+    let disabled = $derived(variant === "disabled");
+    let showLabel = $derived(labelText.length > 0);
+</script>
+
+<div class="relative flex flex-col gap-1 {twMerge(disabled && 'opacity-40')}">
+    {#if showLabel}
+        <label
+            for="textarea"
+            class={twMerge(
+                "text-Content text absolute -top-2 left-2.5 bg-white px-1 text-xs",
```
subiabre: Color names are defined all lowercase in CSS: `text-content` not `text-Content`. This code might work only in environments where the CSS engine performs the styles matching with a case-insensitive behaviour.

src/components/library/TextArea.svelte
```diff
@@ -0,0 +1,65 @@
...
+        {placeholder}
+        {disabled}
+        bind:value
+        class={twMerge(
+            "border-secondary flex h-30 w-110 resize-none items-start gap-2 rounded-lg border bg-white p-3 text-sm text-gray-400 transition-all outline-none",
+            variant ? variantStyles[variant] : "",
+            disabled && "cursor-not-allowed",
+            error && "border-tertiary placeholder:text-tertiary",
+
+            classes,
```
subiabre: Why the empty line between lines 55 and 56?

src/components/library/TextArea.svelte
```diff
@@ -0,0 +1,65 @@
...
+        )}
+    ></textarea>
+
+    {#if helperText}
+        <span class={twMerge("mt-1 ml-4 text-xs", error ? "text-tertiary" : "text-gray-400")}>
```
subiabre: This ternary could be expressed like the `disabled` and `error` expressions a few lines above.

```js
twMerge("mt-1 ml-4 text-xs text-gray-400", error && "text-tertiary");
```

This way we avoid feeding 2 arguments to `twMerge` everytime and we save an unnecessary computation.

subiabre (changes requested):

src/components/library/TextArea.svelte
```diff
@@ -0,0 +1,64 @@
...
+        twMerge(
+            "absolute -top-2.5 left-2.5 bg-white px-1 text-xs transition-colors font-medium",
+            "text-gray-400",
+            error && "text-tertiary",
+            value && !error && "text-content",
+            disabled && "opacity-70",
+        ),
+    );
+
+    let helperClasses = $derived(twMerge("mt-1 ml-3 text-xs text-tertiary transition-all"));
```
subiabre: `twMerge` is never computing anything here. Neither is the `$derived` rune because there are no state-dependent values referenced in here. These classes could simply go in the `class` attribute of the `span` tag for the helper.

subiabre (changes requested):

src/components/library/TextArea.svelte
```diff
@@ -0,0 +1,64 @@
...
+        placeholder,
+        class: className,
+        rows = 4,
+        ...rest
+    }: Props = $props();
+
+    let textareaClasses = $derived(
+        twMerge(
+            "border-secondary flex w-full min-h-[120px] resize-none items-start gap-2 rounded-lg border bg-white p-3 text-sm text-content transition-all outline-none focus:ring-secondary",
+            "placeholder:text-gray-400",
```
subiabre: Why is this in a separate argument? It could be in the string of the first argument and reduce the `twMerge` calculations.

subiabre (changes requested):

subiabre (approved):

PR #226, opened 2026-03-25
Feat: implement component range slider

pepvillalonga:


src/components/library/RangeSlider.svelte
```diff
@@ -0,0 +1,51 @@
...
+        min?: number;
+        max?: number;
+        values?: number[];
+    }
+
+    let {
+        class: classes = "",
+        min = 0,
+        max = 50000,
+        values = $bindable([17000, 34000]),
```
subiabre: Why these arbitrary default values? `max` on 50000? `values` on 17000 and 34000?

src/components/library/RangeSlider.svelte
```diff
@@ -0,0 +1,51 @@
...
+    }: Props = $props();
+
+    const formatCurrency = (value: number) => {
```
subiabre: We already a function `formatCurrency` (in `src/utils/currencies.ts`) which does exactly this but better.

src/components/library/RangeSlider.svelte
```diff
@@ -0,0 +1,51 @@
...
+        return `${value.toLocaleString("es-ES")}€`;
```
subiabre: The application must work with currencies other than the euro and with localizations other than spanish. Do never hardcode these values yourself. Use the existing `formatCurrency` function.

subiabre (changes requested):

subiabre (approved):

PR #228, opened 2026-03-27
Feat: blog template

pepvillalonga:


src/components/library/Share/Iframe.svelte
subiabre, line 1: The convention we follow for svelte components is PascalCase, so this file must start with uppercase I. Also, the name contains a typo: it's `iframe`.

src/components/library/Share/CopyUrl.svelte
```diff
@@ -0,0 +1,18 @@
...
+</script>
+
+<Tooltip
+    text={$t("contributions.tootip.copied")}
+    tooltipClass="bg-secondary -translate-x-[50%]"
+    className="cursor-copy shrink-0 w-auto h-auto"
+>
+    <button
+        class="flex h-23 w-23 items-center justify-center rounded-2xl shadow-[0_1px_3px_0_rgba(0,0,0,0.1)]"
+        onclick={() => navigator.clipboard.writeText(window.location.href)}
```
subiabre: It would be nice if the component took a `url` prop that was the value it copied to the clipboard to allow possible usages where we don't want to copy the current URL.

src/components/library/Share/Facebook.svelte
```diff
@@ -0,0 +1,39 @@
...
+
+        fbWindow = window.open(
+            fbShareUrl,
+            "facebook-share-dialog",
+            `width=${width},height=${height},top=${top},left=${left}`,
+        );
+    }
+
+    onMount(() => {
+        const currentUrl = encodeURIComponent(window.location.href);
```
subiabre: Same as with the `CopyUrl` component. These share components should allow us to pass the URL we want to share, if for example the button to share a project was in a card for the project shown in the home, these components would end up sharing the homepage URL, not the URL to that specific project.

src/components/library/Share/ifreme.svelte
```diff
@@ -0,0 +1,89 @@
...
+    import Loader from "../../../svgs/Loader.svelte";
+
+    interface Props {
+        projectSlug?: string;
+    }
+
+    let { projectSlug = "" }: Props = $props();
+
+    const widgetUrl = $derived(
+        `https://www.goteo.org/widget/project/${projectSlug}`,
```
subiabre: `url` prop.

src/components/library/Share/Facebook.svelte
```diff
@@ -0,0 +1,39 @@
+<script lang="ts">
+    import { onMount } from "svelte";
+
+    import FacebookIcon from "../../../svgs/FacebookIcon.svelte";
+
+    const appId = "184483011630708";
```
subiabre: This `appId` value is ours and identifies Goteo as an app in Meta's 3rd parties registry. Our code should allow for others to use their own Meta app to integrate. There is a `PUBLIC_FACEBOOK_APP_ID` env var, use that as a default value instead.

10/10 if this component allowed to configure the appId as a prop so it would be possible to share in facebook with some one-off app much like the `url` prop.

src/components/library/Share/X.svelte
subiabre, line 1: `X` (uppercase)

src/components/library/Share/x.svelte
```diff
@@ -0,0 +1,26 @@
...
+    interface Props {
+        text?: string;
+    }
+
+    let { text = "" }: Props = $props();
+
+    let currentUrl = $state("");
+
+    onMount(() => {
+        currentUrl = encodeURIComponent(window.location.href);
```
subiabre: Make it a prop.

tsconfig.json
```diff
@@ -2,5 +2,7 @@
   "extends": "astro/tsconfigs/strict",
   "include": [".astro/types.d.ts", "src/types", "**/*"],
   "exclude": ["dist"],
-  "compilerOptions": {}
+  "compilerOptions": {
+    "allowArbitraryExtensions": true
```
subiabre: Remove this line.

subiabre (changes requested):

subiabre (approved):

PR #230, opened 2026-03-31
feat: implement AccordionBox

pepvillalonga:


src/components/library/AccordionBox.svelte
```diff
@@ -0,0 +1,51 @@
...
+    import { slide } from "svelte/transition";
+    import { twMerge, type ClassNameValue } from "tailwind-merge";
+
+    import ChevronDown from "../../svgs/ChevronDown.svelte";
+
+    import type { Snippet } from "svelte";
+
+    interface Props {
+        class?: ClassNameValue;
+        label: string;
```
subiabre: The value of this `label` prop is confusing with other components with a `label` prop that render a true `<label>` tag, but this one does not. The value is actually rendered closer to a heading (`h1`, `h2`, etc), while still not being a true heading. Since the name `heading` is already free I would just rename this.

src/components/library/AccordionBox.svelte
```diff
@@ -0,0 +1,51 @@
...
+        aria-expanded={open}
+    >
+        <span class="text-body text-[15px] font-bold lg:text-[16px]">{label}</span>
+        <span class="text-secondary ml-4 shrink-0">
+            <ChevronDown rotate={open} />
+        </span>
+    </button>
+    {#if open}
+        <div transition:slide|local={{ duration: 300 }}>
+            <div class="text-content px-6 pt-6 pb-6 text-[13px] leading-relaxed lg:text-[14px]">
```
subiabre: Do not hardcode size values. Use a tailwind size class, the one that gets close to the pixels defined in figma is good enough.

subiabre (changes requested):

subiabre (approved):

PR #231, opened 2026-04-01
feat: component HomeBanner

angeldipinto:
I implement the componente `HomeBanner.svelte` from the legacy astro version of it.


src/components/library/ComponentHomeBanner.svelte
subiabre, line 1: No other Svelte component in our source code is prefixed `Component` in their name. Remove it.

src/components/library/ComponentHomeBanner.svelte
```diff
@@ -0,0 +1,82 @@
...
+    });
+
+    const handleClose = () => {
+        isVisible = false;
+        localStorage.setItem(storageKey, "true");
+    };
+</script>
+
+{#if isVisible}
+    <section class="wrapper">
```
subiabre: `wrapper` is an utilitary class of ours to define the common max size of layouts. Components should be able to grow or shrink much more flexibly to allow other layouts to shape them as they please.

Also the inclusion of a `section` tag is problematic as the usage of the component might break the HTML guidelines on the tag: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section (specially because this section lacks a heading, or because the component might be used outside of a context where `section` does not make sense).

subiabre (changes requested):

src/components/home/RenovationBanner.astro
subiabre, line 1: You removed this component but did not replace the existing references to it in the codebase, that is why the Cypress pipeline is failing. This component was referenced in `src/pages/[...lang]/index.astro`, please update it to use your new component.

subiabre (changes requested):

subiabre (approved):

PR #232, opened 2026-04-01
Feat: implement icons

pepvillalonga:


subiabre (changes requested):
Please, review the component paths and names.

We prefer to name our Svelte components with PascalCase filenames. This convention helps them to be easily distinguishable from standard HTML tags when used side to side in front-end code and aligns us with the community where it is also a very wide-spread practice. Please, rename components like `edit.svelte` or `image.svelte` to be PascalCase.

Component names, much like the rest of our codebase, must be named in english. Please, rename components like `Hamburguesa.svelte`, go ahead and also rename the definition for these in the Figma spec file if the spanish name came from there. This allows for easier collaboration world-wide in both development and non-development contexts as we are open-source and work in international settings.

Also check for typos in components like `ifreame.svelte` (iframe) or `matchfaunding.svelte` (match funding) and **very important**: replace special characters in names like `more&less.svelte` for words and letters ("and" in this case), as special characters in file paths might break import statements in some contexts.

src/components/icons/Align.svelte
```diff
@@ -0,0 +1,69 @@
+<script lang="ts">
+    import { twMerge } from "tailwind-merge";
+
+    interface Props {
+        class?: string;
+        width?: string | number;
+
+        align?: string;
```
subiabre: The `prop` align would be much better typed as an enum of the actual values that make something: `left`, `right` and `center`.

src/components/icons/Align.svelte
```diff
@@ -0,0 +1,69 @@
...
+    <!-- Center -->
+    <svg
+        {width}
+        {height}
+        viewBox="0 0 32 32"
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
+        class={twMerge("text-secondary", classes)}
+    >
+        <path
```
subiabre: Across the 3 alignments there is only 1 actual change in the markup and that is in the `<path>` tags where the interior vector shape changes. This code could be minimized incredibly if the component had the `if align ==` logic at the point where things actually change.

src/components/icons/Align.svelte
```diff
@@ -0,0 +1,69 @@
+<script lang="ts">
+    import { twMerge } from "tailwind-merge";
+
+    interface Props {
+        class?: string;
```
subiabre: Remember the better type `ClassNameValues`.

src/components/icons/APIKey.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
+    import { twMerge } from "tailwind-merge";
+
+    interface Props {
+        class?: string;
```
subiabre: Type `ClassNameValues`

src/components/icons/Arrow.svelte
```diff
@@ -0,0 +1,42 @@
...
+{:else}
+    <svg
+        {width}
+        {height}
+        viewBox="0 0 24 24"
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
+        class={twMerge("text-secondary", classes)}
+    >
+        <path
```
subiabre: Again, inside this component there is only a `<path>` change, the outher `svg`  does not change.

src/components/icons/Bag.svelte
```diff
@@ -0,0 +1,38 @@
+<script lang="ts">
+    import { twMerge } from "tailwind-merge";
+    interface Props {
+        class?: string;
+        width?: string | number;
+        height?: string | number;
+        count?: number;
```
subiabre: This icon is heavily tied to our cart/checkout logic (see `src/stores/cart.ts` for example), where this number is mostly referred as `items`. Please update the name of this prop for coherence.

src/components/icons/Bookmark.svelte
```diff
@@ -0,0 +1,49 @@
...
+        class?: string;
+        check?: boolean;
+        width?: string | number;
+        height?: string | number;
+    }
+
+    let { class: classes = "", check = false, width = "24", height = "24" }: Props = $props();
+</script>
+
+{#if !check}
```
subiabre: 👍🏼 nice job unifying the 2 components with this `check` flag. However, again, there should only be a `<path>` change inside the `<svg>` body.

src/components/icons/BudgetMark.svelte
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    let { class: classes = "", width = "24", height = "24", variant }: Props = $props();
+
+    const fillColor = $derived(
+        {
+            tarea: "var(--color-secondary)",
+            material: "var(--color-tertiary)",
+            infraestructura: "var(--color-variant2)",
+        }[variant],
```
subiabre: If you made the object holding the keys of the variants you could have the type of the `variant` prop be `keyof typeof <variantsObject>`. It makes it easier to add or delete one variant without having to touch several lines of code, which might not be as necessary in this specific component but it follows the practice seen in other components like `Tag.svelte` or `Toast.svelte`.

src/components/icons/Chevron.svelte
```diff
@@ -0,0 +1,36 @@
...
+        height?: string | number;
+    }
+
+    let { class: classes = "", direction = "right", width = "24", height = "24" }: Props = $props();
+
+    const paths = {
+        left: "M21 6L11 16L21 26",
+        right: "M11 26L21 16L11 6",
+        up: "M26 21L16 11L6 21",
+        down: "M6 11L16 21L26 11",
```
subiabre, line 8: 💯 this is what I am talking about where I say you can reduce code by only changing the internal `<path>` of the `<svg>` tags. Do this for the other components.

src/components/icons/Chevron.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
+    import { twMerge } from "tailwind-merge";
+
+    interface Props {
+        class?: string;
+        direction?: "left" | "right" | "up" | "down";
```
subiabre: This could be typed `keyof typeof paths` and the set of valid values would automatically be picked up from the `paths` object.

src/components/icons/CloseIconMenu.svelte
```diff
@@ -1,8 +1,23 @@
...
-<svg width={size} height={size} viewBox="0 0 40 40" fill="none" xmlns="http://www.w3.org/2000/svg">
+<svg
+    {width}
+    {height}
+    viewBox="0 0 40 40"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+    class={twMerge("text-secondary", classes)}
+>
     <rect width="40" height="40" rx="8" fill="#59E9D3" />
```
subiabre: `#59e9d3` is the hex value currently defined as the color `primary` in our tailwind theme. Use the name of the color.

src/components/icons/Error.svelte
```diff
@@ -1,26 +1,26 @@
 <script lang="ts">
-    import { type ClassNameValue } from "tailwind-merge";
+    import { twMerge } from "tailwind-merge";
 
     interface Props {
-        class?: ClassNameValue;
-        iconColor?: string;
+        class?: string;
```
subiabre: Why did you remove the better type in here?

subiabre (changes requested):
Please review the comments. Check for the same problems in other icon components as well, the bad typing of the `class` prop with `string` instead of `ClassNameValue` is very wide-spread, as is the abuse of `if/else` structures for changes just on the vector path of the SVG.

Also fix the merge conflicts.

subiabre (approved):

subiabre (approved):

PR #234, opened 2026-04-08
feat: implement component ChatMessage

pepvillalonga:


src/components/library/ChatMessage.svelte
```diff
@@ -0,0 +1,71 @@
...
+                viewBox="0 0 64 64"
+                fill="none"
+                xmlns="http://www.w3.org/2000/svg"
+            >
+                <rect width="64" height="64" rx="8" fill="var(--color-grey)" />
+                <path
+                    d="M32.3156 14.5034C32.4028 14.4851 32.4119 14.5444 32.418 14.6067C33.8836 19.1802 36.0055 22.1742 38.761 25.2792C41.3975 28.2497 44.4463 31.0628 44.6956 36.2596C44.8912 40.3436 43.108 43.6454 41.2172 45.7C39.1074 47.9939 35.682 49.5727 32.111 49.4974C28.4401 49.4191 25.4353 47.7956 23.3118 45.3929C21.1967 43.0001 19.7432 39.9711 20.038 35.8484C20.4707 29.7889 25.0101 26.8298 27.9164 23.0217C29.8057 20.5438 31.1243 17.9808 32.3156 14.5034ZM24.7441 37.0813C24.7805 39.8472 26.2719 42.3723 28.3256 43.6492C31.2228 45.4499 35.253 44.9885 37.6364 42.6216C38.8944 41.3735 39.7485 39.6314 39.9895 37.0813C36.8081 37.0813 36.0821 37.0341 32.3854 37.0341C32.3854 34.5022 32.4187 31.8123 32.4187 29.2819C27.9983 29.2705 24.6903 32.925 24.7441 37.0813Z"
+                    fill="var(--color-black)"
+                />
+            </svg>
```
subiabre: This SVG and the one shown when `isOwn` is `true` are the same except for color changes. Refactor this into a SVG component which takes color props and use that component instead, then pass one color or the other depending on the value of `isOwn`.

src/components/library/ChatMessage.svelte
```diff
@@ -0,0 +1,71 @@
...
+        {/if}
+    </div>
+
+    <!-- Mensaje -->
```
subiabre: Source code, including comments must be in english.

src/components/library/ChatMessage.svelte
```diff
@@ -0,0 +1,71 @@
+<script lang="ts">
+    import { twMerge, type ClassNameValue } from "tailwind-merge";
+
+    interface Props {
+        type: "own" | "foreign";
+        name: string;
+        message: string;
+        date: string;
```
subiabre: Use a proper `Date` type and then use the function `formatDate` from `src/utils/dates.ts` to print the date in a format appropriate to the user.

Almost never you want to work with dates as a `string` in any language. Just the amount of different formats used worldwide would make any format you choose for your string dates invalid for a huge number of people that prefer other formats, there are also precision problems regarding timezones and daylight savings and arithmetic problems when sorting dates and the fact that you can't add or substract any time from a string.

That and much more is usually just handled in the background by the language for you if you simply use the appropriate type. Please use them.

subiabre (changes requested):

subiabre (approved):

PR #236, opened 2026-04-09
fix: fix variables and remove unused colors

pepvillalonga:
Now I need to change the variables in all the files that use Tailwind (I've noted down the ones I've changed so I can look them up.)


subiabre (changes requested):
Code replacements look thorough and sufficient.

However check the pipelines. Both are failing. I can confirm the cypress issue is not related to changes on this branch, but there are formatting issues that you must resolve.

subiabre (approved):

PR #239, opened 2026-04-14
Feat: implement component chat textarea

pepvillalonga:


src/components/library/ChatTextarea.svelte
```diff
@@ -0,0 +1,35 @@
...
+
+    import { t } from "../../i18n/store";
+    import SendIcon from "../icons/SendIcon.svelte";
+
+    interface Props {
+        class?: ClassNameValue;
+        value?: string;
+        name?: string;
+        id?: string;
+        onsubmit?: (value: string) => void;
```
subiabre: The convention on the rest of the components is to have our callback props use camelCase.

src/components/library/ChatTextarea.svelte
```diff
@@ -0,0 +1,35 @@
...
+    import { twMerge, type ClassNameValue } from "tailwind-merge";
+
+    import { t } from "../../i18n/store";
+    import SendIcon from "../icons/SendIcon.svelte";
+
+    interface Props {
+        class?: ClassNameValue;
+        value?: string;
+        name?: string;
+        id?: string;
```
subiabre: The textarea [name](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/textarea#name) (and also the id attribute) are focused on providing "control" functionality. But we already control the submission of the internal textarea and capture its value with the callback prop.

Having two different control layers able to modify the same thing at the same thing can (and already has lead in this project) to very hard to debug issues. Please remove these attributes.

subiabre (changes requested):

subiabre (approved):

PR #241, opened 2026-04-14
[feat] Dynamic Toggle component

Dayv1dDev:


src/components/library/Toggle.svelte
```diff
@@ -0,0 +1,91 @@
...
+    }: {
+        onChange: (newState: boolean) => void;
+        colorsOn?: { bg: string; circle: string };
+        colorsOff?: { bg: string; circle: string };
+        padding?: number;
+        width?: number;
+        height?: number;
+    } = $props();
+
+    // true = on ----- false = off
```
subiabre: I would not have it any other way! wtf

src/components/library/Toggle.svelte
```diff
@@ -0,0 +1,91 @@
...
+    let toggle = $state(false);
+
+    const getStyleOrClass = (value: string) => {
+        const isTailwind =
+            !value.startsWith("#") && !value.startsWith("rgb") && !value.startsWith("hsl");
+        return {
+            class: isTailwind ? value : "",
+            style: isTailwind ? "" : value,
+        };
+    };
```
subiabre: Unnecessarily complex. We already avoid hardcoding hex values of colours across the codebase, if any future developer wishes to use hex values with this component they can do so by using tailwind's syntax for arbitrary values: https://tailwindcss.com/docs/styling-with-utility-classes#using-arbitrary-values

You can remove the entire `getStyleOrClass` thing by just comforming to tailwind.

src/components/library/Toggle.svelte
```diff
@@ -0,0 +1,91 @@
...
+    function handleToggle() {
+        toggle = !toggle;
+        if (onChange) onChange(toggle);
+    }
+
+    $effect(() => {
+        // Ensure width is double the height for proper toggle functionality
+        if (width !== 2 * height) {
+            console.warn(
+                `Width should be double the height for proper toggle functionality. Current width: ${width}px, height: ${height}px. Please adjust accordingly.`,
```
subiabre: You could save computing and still warn fellow developers better by removing this `$effect` rune and adding this warn as a doc block comment on the `width` prop.

subiabre (changes requested):

PR #242, opened 2026-04-15
Feat: implement component actionable button

pepvillalonga:


src/components/library/ActionableButton.svelte
```diff
@@ -0,0 +1,123 @@
...
+                <polyline points="20 6 9 17 4 12"></polyline>
+            </svg>
+        {/if}
+    {:else}
+        {@render children()}
+    {/if}
+</button>
+
+<style>
+    .spinner {
```
subiabre: Why not use the already existing `Spinner` Svelte component?

src/components/library/ActionableButton.svelte
```diff
@@ -0,0 +1,123 @@
...
+            }
+        }
+    }
+
+    export function reset() {
+        state = "accionable";
+    }
+</script>
+
+<button
```
subiabre: Why not use the already existing `Button` svelte component?

src/components/library/ActionableButton.svelte
```diff
@@ -0,0 +1,123 @@
+<script module lang="ts">
+    export type ActionableState = "accionable" | "accionando" | "accionado";
+</script>
```
subiabre: This is highly unnecessary, you can just define the type for the state of the button as a type in the `script` tag of the svelte component and give it to the `state` prop. Where would you need to use the possible states of this actionable button without the button?

Also, remember to use english.

subiabre (changes requested):

subiabre (approved):

PR #245, opened 2026-04-16
Feat/add remaining filters to search box

angeldipinto:
The 'Territory' search filter and categories have been added.

---
  - https://app.asana.com/0/0/1213854146709593


src/components/search/CategoryFilter.svelte
```diff
@@ -4,63 +4,91 @@ Interactive category selection using existing categories from utils/categories.t
...
-    async function getAvailableCategories(): Promise<Category[]> {
-        const { data } = await apiCategoriesGetCollection();
+    onMount(async () => {
+        try {
+            const response = await apiCategoriesGetCollection();
 
-        if (!data) {
-            return [];
+            const responseData = response as any;
+            const rawData = responseData.data || responseData["hydra:member"] || responseData;
```
subiabre: The object returned by the SDK function `apiCategoriesGetCollection` (and every other read function from the SDK) returns a shape already documented.

Here you are overwriting the type of the response as `any` (I suppose so that the IDE does not complain when you try to access the non-existing keys of the response object `hydra:member`, which will only come when you make the request with a `Accept: application/ld+json` HTTP header). Overwriting the type of a returned type is rarely-ever a good solution.

Simply check the `.data` and `.error` keys of the response object, or destructure them like:
```js
const { data, error } = await <sdkFunction>;
```

src/components/search/CategoryFilter.svelte
```diff
@@ -4,63 +4,91 @@ Interactive category selection using existing categories from utils/categories.t
...
+
+            if (Array.isArray(rawData)) {
```
subiabre: Unnecessary `if` sentence, the SDK already tells us the type of the returned data is an array.

src/components/search/CategoryFilter.svelte
```diff
@@ -4,63 +4,91 @@ Interactive category selection using existing categories from utils/categories.t
...
+                categories = rawData.map((cat: any) => ({
+                    ...cat,
+                    id: String(cat.id),
+                    name: cat.name || cat.title || "Category",
```
subiabre: None of these properties exist on the returned `Category` objects that come from the API. If you insist on your local `Category` objects on having a key `name` then you must search for other ways to get that name.

src/components/search/CategoryFilter.svelte
```diff
@@ -4,63 +4,91 @@ Interactive category selection using existing categories from utils/categories.t
...
+            ? selectedCategories.filter((id) => id !== categoryId)
+            : [...selectedCategories, categoryId];
 
-    function mapCategoryToOption(category: Category): Option {
-        return {
-            id: category.id!,
-            text: $t("categories." + category.id!),
-        };
+        if (onCategoryChange) {
+            onCategoryChange(newSelection);
```
subiabre: You can use the `?` operator to conditionally call this function if it exists in a single line:

```js
onCategoryChange?.(newSelection);
```

src/components/search/CategoryFilter.svelte
```diff
@@ -4,63 +4,91 @@ Interactive category selection using existing categories from utils/categories.t
...
-        <CategorySelect
-            bind:selected
-            selectedIds={selected.map((s) => s.id)}
-            options={categories.map((c) => mapCategoryToOption(c))}
-            onchange={(selected) => onCategoryChange?.(selected.map((o) => `${o.id}`))}
-        />
-    {/await}
+<div class="flex flex-col gap-4" data-testid={testId}>
+    <h3 class="text-sm font-bold tracking-wider text-gray-500 uppercase">
+        {$t("search.categoryLabel")}
```
subiabre: Move this key to one of the groups: common, domain, pages or system.

src/components/search/CategoryFilter.svelte
```diff
@@ -4,63 +4,91 @@ Interactive category selection using existing categories from utils/categories.t
...
+    {:else if categories.length > 0}
+        <div class="flex flex-wrap gap-2">
+            {#each categories as category}
+                {@const isSelected = selectedCategories.includes(category.id)}
+                <button
+                    type="button"
+                    onclick={() => toggleCategory(category.id)}
+                    class="rounded-full border px-4 py-2 text-sm font-medium transition-all {isSelected
+                        ? 'border-purple-800 bg-purple-800 text-white shadow-sm'
+                        : 'border-gray-300 bg-white text-gray-600 hover:border-purple-400 hover:text-purple-600'}"
```
subiabre: A component already exists for this: `Category` from "src/components/library/Category.svelte". Please use the existing components every time you can, avoid re-implementing things.

src/components/search/CategoryFilter.svelte
```diff
@@ -4,63 +4,91 @@ Interactive category selection using existing categories from utils/categories.t
...
+        {$t("search.categoryLabel")}
+    </h3>
 
-    {#if selected.length > 0}
-        <div class="mt-4 text-sm text-[#3d3d3d] opacity-70">
-            {$t("search.selectedCategories", { count: selected.length })}
+    {#if isLoading}
+        <div class="flex flex-wrap gap-2">
+            {#each Array(6) as _}
+                <div class="h-10 w-24 animate-pulse rounded-full bg-gray-100"></div>
```
subiabre, line 47: This is good, it would be even better if this existed as an "skeleton" variant of the `Category` svelte component (without the arbitrary sizes and respecting the actual shape on the component).

src/components/search/CategoryFilter.svelte
```diff
@@ -4,63 +4,91 @@ Interactive category selection using existing categories from utils/categories.t
...
-    {/await}
+<div class="flex flex-col gap-4" data-testid={testId}>
+    <h3 class="text-sm font-bold tracking-wider text-gray-500 uppercase">
+        {$t("search.categoryLabel")}
+    </h3>
 
-    {#if selected.length > 0}
-        <div class="mt-4 text-sm text-[#3d3d3d] opacity-70">
-            {$t("search.selectedCategories", { count: selected.length })}
+    {#if isLoading}
```
subiabre: You can avoid this `isLoading` internal state logic by simply awaiting for the promise in the Svelte markup: https://svelte.dev/docs/svelte/await

Take this example:

```ts
<script lang="ts">
const categories = apiCategoriesCollectionGet();
</script>

{#await expression}
    // is loading block here
{:then name}
    // loaded categories block here
{:catch name}
    // error during request block here
{/await}
```

src/components/search/SearchFilters.svelte
```diff
@@ -11,16 +11,25 @@ Integrated with searchStore for state management and URL synchronization
...
         locale?: string;
         initialFilters?: {
             query?: string;
             statusFilter?: string;
             categories?: string[];
+            territory?: {
+                country?: string | null;
+                subLvl1?: string | null;
+                subLvl2?: string | null;
+                rawQuery?: string;
```
subiabre: This is bad. If the `Territory` shape changed in the API this would break without you even knowing why. What if the API changes the name of one of their internal keys: `subLvl1` -> `region`? What if the API drops one of the keys or adds a key? Your filter would be out of sync. That's why we have an SDK that can be automatically generated from the API spec, and that is why we prefer to use the types from the SDK. You should always avoid re-implementing as much as you can.

If you need a type that depends on another type you can use inheritance or composition to create a new type that better suits your needs.

Inheritance example:
```ts
const MyCustomTerritoryExample extends Territory {
    rawQuery: string;
}

const territory = new MyCustomTerritoryExample(); // {  rawQuery: "", country: "", ... }
```

Composition example:
```ts
const MyCustomTerritoryExample {
    rawQuery: string;
    territory: Territory;
}

const territory = new MyCustomTerritoryExample(); // {  rawQuery: "", territory: { country: "", ... } }
```

src/components/search/SearchFilters.svelte
```diff
@@ -33,23 +42,27 @@ Integrated with searchStore for state management and URL synchronization
...
         if (initialFilters) {
             searchStore.initializeFilters({
                 query: initialFilters.query || "",
                 statusFilter: initialFilters.statusFilter || "",
                 categories: initialFilters.categories || [],
+                territory: {
+                    country: initialFilters.territory?.country || null,
+                    subLvl1: initialFilters.territory?.subLvl1 || null,
+                    subLvl2: initialFilters.territory?.subLvl2 || null,
+                },
```
subiabre: Again, this implementation will break when the shape of the API-given `Territory` type changes. Simply use the `initialFilters.territory` object. If you need to re-assign the keys, use [destructure](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring), but avoid hard-coding and repeating stuff as much as you can.

src/components/search/SearchFilters.svelte
```diff
@@ -33,23 +42,27 @@ Integrated with searchStore for state management and URL synchronization
...
     // Dropdown options - use translation keys for labels
     const statusOptions = [
         { value: "all", translationKey: "filters.status.all" },
-        // API status values - pass directly to backend
-        { value: "in_campaign", translationKey: "filters.status.funding" }, // Actively raising funds
-        { value: "in_funding", translationKey: "filters.status.successful" }, // Successfully raised, receiving funds
-        { value: "funded", translationKey: "filters.status.completed" }, // Completed funding process
+        { value: "in_campaign", translationKey: "filters.status.funding" },
+        { value: "in_funding", translationKey: "filters.status.successful" },
+        { value: "funded", translationKey: "filters.status.completed" },
```
subiabre: Move these translations keys to keys inside "domain.search.filters.status".

src/components/search/SearchFilters.svelte
```diff
@@ -103,35 +124,45 @@ Integrated with searchStore for state management and URL synchronization
...
-                <FilterIcon width="16" height="16" class="mr-2" />
-                {$t("search.showFilters")}
-            </SearchButton>
-        {/if}
+        <SearchButton
+            variant="ghost"
+            onclick={toggleFilters}
+            data-testid="toggle-filters"
+            aria-expanded={filtersOpen}
+            class="hover:bg-grey w-full justify-center rounded-xl border border-[#e5e5e0] bg-transparent px-4 py-2 transition-colors min-[500px]:w-auto"
```
subiabre: Why is it necessary to hard-code the min width to `500px`? Why is it necessary to pass these custom classes to the search button?

src/components/search/TerritoryInputFilter.svelte
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    let {
+        value = $bindable(""),
+        label = "",
+        placeholder = "",
+        onTerritoryDetected,
+    }: Props = $props();
+
+    const inputId = `territory-input-${Math.random().toString(36).slice(2, 9)}`;
```
subiabre: Use Svelte's `$props.id` function to generate IDs for the markup inside your Svelte code.

Moreover, if the `$props.id` function was not available this implementation is not good. You would have needed to use the `hash` function or something else than simply having faith in randomness.

src/components/search/TerritoryInputFilter.svelte
```diff
@@ -0,0 +1,62 @@
...
+
+    async function handleValidation(text: string) {
+        // Si el campo está vacío, reseteamos los filtros de territorio en el store
```
subiabre: In english please.

src/components/search/TerritoryInputFilter.svelte
```diff
@@ -0,0 +1,62 @@
...
+        if (!text || text.trim().length < 2) {
+            return;
+        }
+
+        // Llamada al servicio fetch de Nominatim
+        const result = await nominatim.findIsoByText(text);
+
+        // Enviamos los datos encontrados (o nulls si no hubo match) al padre
+        onTerritoryDetected({
+            rawQuery: text || "",
```
subiabre: Right-hand side of this assignation will never execute because the checks in line 30 prevent any value in the `text` variable to ever evaluate as falsey.

src/components/search/TerritoryInputFilter.svelte
```diff
@@ -0,0 +1,62 @@
...
+
+    <SearchInput
+        id={inputId}
+        {value}
+        {placeholder}
+        label=""
+        onSearch={(text) => (value = text)}
+        onEnter={() => handleValidation(value)}
+        onBlur={() => handleValidation(value)}
+        data-testid="territory-search-input"
```
subiabre: One of our developers added these attributes because he was using an external tool to test the inputs he was implementing and needed this attribute. If you do not need these do not add them.

src/pages/[...lang]/search.astro
```diff
@@ -13,24 +13,35 @@ const initialProjects: Project[] = [];
 
 // Parse URL search parameters for SSR
 const url = new URL(Astro.request.url);
+
+/**
+ * se agrega  la lógica para extraer los datos geográficos de la URL
+ * Esto permite que al compartir el link, la ubicación se mantenga.
```
Comment (subiabre) [inline src/pages/[...lang]/search.astro]: If your doc block comment is about the territory changes add them on the key of the `initialFilters` object that is about the territory changes you've made.

src/pages/[...lang]/search.astro
```diff
@@ -13,24 +13,35 @@ const initialProjects: Project[] = [];
...
+ */
 const initialFilters = {
     query: url.searchParams.get("q") || "",
     statusFilter: url.searchParams.get("status") || "",
     categories: url.searchParams.get("categories")?.split(",").filter(Boolean) || [],
+    territory: {
+        country: url.searchParams.get("country") || null,
+        subLvl1: url.searchParams.get("subLvl1") || null,
+        subLvl2: url.searchParams.get("subLvl2") || null,
+        rawQuery: url.searchParams.get("locationText") || "",
```
Comment (subiabre) [inline src/pages/[...lang]/search.astro]: Use destructuring.

src/services/nominatim.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import type { Territory } from "../openapi/client";
+
+export const nominatim = {
+    /**
+     * Searches for a location and returns ISO codes
+     * @param text The location string to search for
+     */
+    async findIsoByText(text: string): Promise<Territory | null> {
+        if (!text || text.trim().length < 3) return null;
```
subiabre: This prevents users of this service to search for valid places like "Ulm" (Germany), "Goa" (India), "Ica" (Perú). If you want to prevent hitting the Nominatim service unnecesarily you can validate the given text is at least 1 character long when trimmed (which would allow to search for even a random place in scandinavia with the shortest name ever).

src/services/nominatim.ts
```diff
@@ -0,0 +1,54 @@
...
+        try {
+            // We use addressdetails=1 to get ISO codes in the response
+            const params = new URLSearchParams({
+                q: text,
+                format: "json",
+                addressdetails: "1",
+                limit: "1",
+            });
+
+            const response = await fetch("https://nominatim.openstreetmap.org/search?" + params, {
```
subiabre: Use the native browser [cache API](https://developer.mozilla.org/en-US/docs/Web/API/Cache) to reduce the usage of nominatim.

src/services/nominatim.ts
```diff
@@ -0,0 +1,54 @@
...
+                headers: {
+                    // Nominatim requires a descriptive User-Agent
+                    "User-Agent": "BuscarProyecto(contact@yourproyect.com)",
```
subiabre: Why the random non-existing email address in the user agent? The user agent shall be a descriptive string of _who_ is hitting the site, this tells nominatim nothing about us. When the API needs to make HTTP requests to other service the user agent we use is `Platoniq Goteo-v4-API`, maybe the front should use `Platoniq Goteo-v4-web`?

src/services/nominatim.ts
```diff
@@ -0,0 +1,54 @@
...
+
+            const data = await response.json();
+
+            if (data && data.length > 0) {
+                const address = data[0].address;
+
+                return {
+                    country: address.country_code?.toUpperCase() || null,
+                    subLvl1: address["ISO3166-2-lvl4"] || address["ISO3166-2-lvl3"] || null,
+                    subLvl2: address["ISO3166-2-lvl6"] || null,
```
subiabre: Could not find much documentation about how many levels Nominatim has for their "ISO3166-2-lvl<number>" keys, but it is clear that there is a range of levels, so we can never trust that our first available sub-division ISO code might be at `ISO3166-2-lvl4` or `ISO3166-2-lvl3`, the same for the ISO code of our second level.

You need to actively search for the highest available levels in the response to obtain the `subLvl` here.

Take this pseudo-code for inspiration:

```
let subLvl1 = null;
for (level in 1..100) {
    const isoLevel = "ISO3166-2-lvl" . level;
    if (!res[isoLevel]) {
        continue; // no ISO code found at the current level
    }

    subLvl1 = res[isoLevel];
    return; // ISO code found at the current level, set code in var and stop loop
}
```

subiabre (changes requested):
Additional to the given review comments, check all your changes and replace the comments you wrote in spanish throughout the code and write them in english.

PR #249, opened 2026-04-21
Feat/replace hash function cyrb53 with murmurhash3

angeldipinto:
The cyrb53 function has been replaced by murmurhash3. This is better and more maintainable.

---
  - https://app.asana.com/0/0/1214016556815497


subiabre (changes requested):
Code looks good, nice touch updating the Svelte IDs with `$derived` stores. But there are code style issues.

Do not forget to check the pipelines. Currently we know the cypress pipeline is broken until we recover the register page, but the prettier pipeline needs to be green everytime. Develop the habit of running the `format` command defined in our `package.json` before every commit. You can look into [git hooks](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) if you want to make this automatically on your local machine.

subiabre (approved):

PR #253, opened 2026-04-23
Fix: translation keys

pepvillalonga:
---
  - https://app.asana.com/0/0/1213980243316528


src/i18n/locales/es.json
```diff
@@ -208,6 +282,9 @@
       "file": {
         "sizeTooLarge": "El archivo {{ file }} supera el tamaño máximo ({{ max }})",
         "unsupportedType": "Tipo de archivo no permitido: {{ file }}"
+      },
+      "lang": {
+        "notAvailable": "El idioma seleccionado \"{{lang}}\" no está disponible para este proyecto. Mostrando contenido en el idioma disponible."
```
subiabre: Given how this message is highly coupled with the project page (in facts that's the only place where this message is displayed) I would move this key to something like `pages.project.view.langNotAvailable`.

I would not expect a `system.lang.notAvailable` key to give me a message that only makes sense in a project page.

subiabre (changes requested):

src/i18n/locales/es.json
```diff
@@ -222,49 +225,80 @@
             "type": "El tipo de partida se debe especificar",
             "class": "La clase de partida se debe especificar"
           }
-        },
-        "aboutYou": {
```
subiabre, line 226: All of these keys are deleted without replacement, but they are necessary for the "about you" tab of the project edit form.

subiabre (changes requested):

subiabre (approved):

PR #257, opened 2026-04-28
Fix/translation keys pages.search

pepvillalonga:
---
  - https://app.asana.com/0/0/1214301600008669

### DIFF
--- a/src/i18n/locales/es.json
+++ b/src/i18n/locales/es.json
@@ -49,7 +49,24 @@
           "multiple": "Hay {{ units }} unidades"
         },
         "isInfinite": "Sin límite de unidades"
+      },
+      "status": {
+        "all": "Todos los estados",
+        "funding": "En financiación",
+        "successful": "Exitosas",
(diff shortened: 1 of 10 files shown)


src/i18n/locales/es.json
```diff
@@ -49,7 +49,24 @@
           "multiple": "Hay {{ units }} unidades"
         },
         "isInfinite": "Sin límite de unidades"
+      },
+      "status": {
```
subiabre, line 53: These statuses relate only to a project. Given how there are other resources from the API such as `GatewayCheckout` that also have a status I would move this key inside `domain.project.status` to avoid future confusion with other resource statuses.

subiabre (changes requested):

src/i18n/locales/es.json
pepvillalonga: There was no mistake.

subiabre (approved):

PR #265, opened 2026-05-04
Feat: implement submenu functionality in header

pepvillalonga:
---
  - https://app.asana.com/0/0/1214492339678082


subiabre (changes requested):
While the header item expands as requested and mantains a coherent backdrop blur, when expanded it pushes down the items behind it. I remember an early iteration of this component that only pushed the items when at the top of the scroll, what happened to it? That behaviour was ideal, feels very weird to be at halfway of the page and have the whole site be pushed down when I open the floating header hamburger icon

subiabre (approved):

PR #269, opened 2026-05-08
feat: cart.ts-keys

pepvillalonga:
---
  - https://app.asana.com/0/0/1214492339678078

### DIFF
--- a/src/stores/cart.ts
+++ b/src/stores/cart.ts
@@ -9,6 +9,11 @@ export interface CartItem extends GatewayCharge {
     kind: "free" | "reward" | "tip";
     quantity: number;
 
+    /**
+     * `target` references the Accounting that will receive the money\
+     * `recipient` references the owner of that Accounting
+     */
+    recipient: string;
     recipientDisplayName: string;
 
     /**
@@ -21,12 +26,10 @@ type CartState = {
     items: Record<string, CartItem>;
 };
 
-export type CartItemInput = Omit<CartItem, "key"> & { recipient: string };
-
 export interface CartStore {
     subscribe: (run: (value: CartState) => void) => () => void;
 
-    addItem: (item: CartItemInput) => void;
+    addItem: (item: Omit<CartItem, "key">) => void;
     removeItem: (key: string) => void;
     updateQuantity: (key: string, quantity: number) => void;
 
@@ -36,7 +39,7 @@ export interface CartStore {
 
 const isBrowser = typeof window !== "undefined";
 
-function generateKey(item: CartItemInput): string {
+function generateKey(item: Omit<CartItem, "key">): string {
     const base = `${item.kind}:${item.recipient}`;
     if (item.kind === "reward" && item.reward?.id != null) {
         return `${base};reward:${apiProjectRewardsIdGetUrl.replace("{id}", String(item.reward.id))}`;
@@ -90,7 +93,7 @@ function createCartStore(): CartStore {
     return {
         subscribe,
 
-        addItem: (item: CartItemInput) =>
+        addItem: (item: Omit<CartItem, "key">) =>
             update((cart) => {
                 const key = generateKey(item);
                 const items = { ...cart.items };
@@ -178,4 +181,3 @@ export const cartByRecipient = derived(cart, ($cart) => {
 
     return grouped;
 });
-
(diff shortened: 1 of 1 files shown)


src/stores/cart.ts
```diff
@@ -25,31 +23,36 @@ type CartState = {
...
-};
-
 const isBrowser = typeof window !== "undefined";
 
-function generateKey(args: GenerateKeyOptions): string {
-    return murmur
-        .murmur3(`${args.kind}${args.target}-${args.title}${args.description}-${args.dateCreated}`)
-        .toString(16);
+function generateKey(kind: CartItem["kind"], recipient: string, reward?: ProjectReward): string {
+    const prefix = kind === "tip" ? "tip" : "project";
```
subiabre: This prefix strategy does not scale. Why not simply use the `kind` for the prefix? We want to include more recipients to Goteo in the future and our current cart management should allow us to do it as painless as possible.

src/stores/cart.ts
```diff
@@ -25,31 +23,36 @@ type CartState = {
...
+    const base = `${prefix}:${recipient}`;
+    if (kind === "reward" && reward?.id != null) {
+        return `${base};reward:/v4/project_rewards/${reward.id}`;
```
subiabre: Avoid hardcoding as much as possible. While it is unlikely that the structure of the IRIs in the API changes, this should be done using the SDK: https://heyapi.dev/openapi-ts/clients/fetch#build-url

Our SDK also has a plugin which makes it export all the paths in constants of the same name as their operation with the `Url` suffix. In this case: `apiProjectRewardsIdGetUrl` for the URL of a single project reward item.

src/stores/cart.ts
```diff
@@ -12,7 +11,6 @@ export interface CartItem extends GatewayCharge {
      * `target` references the Accounting that will receive the money\
      * `recipient` references the owner of that Accounting
      */
-    recipient: string;
```
subiabre, line 15: You removed this line but not the docblock comment above it that documented the difference between `recipient` and `target`.

src/stores/cart.ts
```diff
@@ -25,31 +23,36 @@ type CartState = {
...
 const isBrowser = typeof window !== "undefined";
 
-function generateKey(args: GenerateKeyOptions): string {
-    return murmur
-        .murmur3(`${args.kind}${args.target}-${args.title}${args.description}-${args.dateCreated}`)
-        .toString(16);
+function generateKey(kind: CartItem["kind"], recipient: string, reward?: ProjectReward): string {
+    const prefix = kind === "tip" ? "tip" : "project";
+    const base = `${prefix}:${recipient}`;
+    if (kind === "reward" && reward?.id != null) {
```
subiabre: The reward is presumed to come in the cart item itself for items of kind "reward". It should use the reward of the item itself. The current function signature can then be stripped of the `reward` argument. If you think this is better you should then update the calls to this function when an item of "reward" kind is added to pass the associated reward.

src/stores/cart.ts
```diff
@@ -25,31 +23,36 @@ type CartState = {
...
+    }
+    return base;
+}
+
+function parseKey(key: string): { recipient: string; rewardIRI?: string } {
+    const segments = key.split(";");
+    const recipient = segments[0].slice(segments[0].indexOf(":") + 1);
+    const rewardSegment = segments.find((s) => s.startsWith("reward:"));
+    const rewardIRI = rewardSegment?.slice("reward:".length);
+    return { recipient, rewardIRI };
```
subiabre: For a original taxonomy of

> `<kind>:<recipient_IRI>[;<extra_kind>:<recipient_IRI>...]`

This parsing does not really reflect it. To avoid loss of data and to make this function support future kinds of cart items the output of this parsing should be an object like:

```
{ kind: string, recipient: string, extra: { kind: string, recipient: string }[] }
```

src/stores/cart.ts
```diff
@@ -163,11 +166,12 @@ export const cartByRecipient = derived(cart, ($cart) => {
...
-        }
+        const { recipient } = parseKey(item.key);
+        grouped[recipient] ??= [];
+        grouped[recipient].push(item);
     }
 
     return grouped;
 });
+
+export { parseKey };
```
subiabre: You could simply add `export` to the function declaration if you want to make this exportable. Which is the de-facto convention across the rest of the codebase.

subiabre (changes requested):

subiabre (approved):

PR #270, opened 2026-05-08
Fix: homebanner correct buttons

pepvillalonga:
---
  - https://app.asana.com/0/0/1214608472905126

### DIFF
--- a/src/components/library/HomeBanner.svelte
+++ b/src/components/library/HomeBanner.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
-    import CloseMenu from "../icons/CloseMenu.svelte";
+    import Button from "./Button.svelte";
+    import Close from "../icons/Close.svelte";
 
     interface Props {
         title: string;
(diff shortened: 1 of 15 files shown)


src/components/library/HomeBanner.svelte
```diff
@@ -21,54 +22,56 @@
...
-            class="relative z-10 flex flex-col gap-4 md:flex-row md:items-center md:justify-between"
-        >
-            <div class="flex-1 space-y-3">
-                <h2 class="text-2xl font-bold tracking-tight sm:text-3xl md:text-4xl">
-                    {title}
-                </h2>
-                <p class="text-variant1 max-w-3xl text-sm opacity-90 sm:text-base md:text-lg">
-                    {description}
-                </p>
+    <div class="wrapper">
```
subiabre: Components should be able to scale at will and adjust to the size of their parents. They can have default sizes, but `wrapper` is an utilitary function intended to limit the max width of our layouts, not individual components, and because it is outside of tailwind realm it cannot be overriden via the props class and `twMerge`

subiabre (changes requested):

subiabre (approved):

PR #273, opened 2026-05-11
[feat] Allow Dynamic Routes with ACL + add project/[idOrSlug]/edit ACL

Dayv1dDev:
---
  - https://app.asana.com/0/0/1214639590504508

### DIFF
--- a/src/firewall/index.ts
+++ b/src/firewall/index.ts
@@ -27,6 +27,12 @@ function getRegex(path: string): RegExp {
     return regexCache.get(path)!;
 }
 
+/**
+ * Checks if a pathname matches a given path.
+ * @param pathname The pathname to check, e.g. "/admin/dashboard"
+ * @param path The path to match against, which can be an exact path (e.g. "/admin") or a regex pattern (e.g. "/admin/.*")
+ * @returns True if the pathname matches the path, false otherwise
+ */
 function matchesPath(pathname: string, path: string): boolean {
     // Exact match
     if (pathname === path) {
--- a/src/firewall/index.ts
+++ b/src/firewall/index.ts
@@ -15,8 +15,13 @@ function normalizePath(pathname: string): string {
     return "/" + segments.join("/");
 }
 
-function isRegexPath(path: string): boolean {
-    return /[.*+?^${}()|[\]\\]/.test(path);
+function isValidRegex(pattern: string): boolean {
+    try {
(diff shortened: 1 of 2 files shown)


src/firewall/index.ts
```diff
@@ -13,8 +15,31 @@ function normalizePath(pathname: string): string {
...
+
+function getRegex(path: string): RegExp {
+    if (!regexCache.has(path)) {
+        regexCache.set(path, new RegExp(`^${path}$`));
+    }
+
+    return regexCache.get(path)!;
+}
+
+function matchesPath(pathname: string, path: string): boolean {
```
subiabre, line 41: Please, add doc-block to explain this function and specially which argument is the one used as truth and which one is compared against the truth.

src/firewall/index.ts
```diff
@@ -13,8 +15,31 @@ function normalizePath(pathname: string): string {
     return "/" + segments.join("/");
 }
 
-function matchesPath(pathname: string, basePath: string): boolean {
-    return pathname === basePath || pathname.startsWith(basePath + "/");
+function isRegexPath(path: string): boolean {
+    return /[.*+?^${}()|[\]\\]/.test(path);
```
subiabre: Will take good-looking regex strings as valid regex, which concerns me because developers might accidentally produce broken regex strings in the ACL that will never match against anything and silently break the firewall. A better approach would be to try and compile the regex string:

```ts
function isValidRegex(pattern: string): boolean {
    try {
        new RegExp(pattern);
        return true;
    } catch {
        return false;
    }
}
```

src/firewall/index.ts
```diff
@@ -31,4 +56,4 @@ export function getMatchingACL(pathname: string): ControlItem | null {
 
 export function isAuthorized(control: ControlItem, roles: string[]): boolean {
     return control.roles.some((role) => roles.includes(role));
```
subiabre, line 79: I know it's not your line, but I think this currently returns always `false` when the `control` has no roles defined. But I think it's normal to assume that when you define and ACL item with no roles it means that anyone can access it, rather than nobody.

Please, check this behaviour, make the appropriate improvements and add a doc-block to explain that `isAuthorized` is true when at least one of the roles is in the control.

subiabre (changes requested):

subiabre (approved):

PR #275, opened 2026-05-12
feat: add browser cache interceptor for OpenAPI SDK client

pepvillalonga:
---
  - https://app.asana.com/0/0/1214492339678072


src/openapi/cacheFetch.ts
```diff
@@ -0,0 +1,21 @@
+export const PLATINIQ_API_CACHE = "platiniq-api";
```
subiabre: This application is an open-core that might not be run by or owned by us. If this cache is for the underlying API, it's name is "the v4 API".  Code still works fine with this name, but in this instance the name actually matters, would be like naming a cache for an Instagram API "meta-api". It's too ambiguous and, in the long run, prone to collision with other services of us we would like to add to Goteo.

Also, it's Plat**o**niq.

src/openapi/api.ts
```diff
@@ -101,5 +102,6 @@ export const createClientConfig: CreateClientConfig = (config) => {
     return {
         ...config,
         baseUrl: getBaseUrl(),
+        fetch: createCachedFetch(PLATINIQ_API_CACHE),
```
subiabre: Very bad. It overrides the underlying `fetch` client, which could not only makes us potentially miss all the goodies of the default Hey API fetch client like default headers, CORS handling or JSON serialization, but also will force every future GET request to be cached by default.

The recommended approach was based on interceptors because they can be attached or ejected from the client at will: https://heyapi.dev/openapi-ts/clients/fetch#interceptors

That allows us to only use cached requests when it's actually possible (by attaching the interceptor to the API client in browser-run code without having to always check if `browser` exists) and necessary (public project campaign pages might not change as often, but serving the same cached response when a user wants to check the status of a payment they made would be very bad).

Please, use an interceptor.

subiabre (changes requested):

subiabre (approved):

PR #281, opened 2026-05-14
[fix] Adjust several styles in CreateProjectForm, matching now with Figma

Dayv1dDev:
---
  - https://app.asana.com/0/0/1214698734280481

### DIFF
--- a/src/components/BaseCard.svelte
+++ b/src/components/BaseCard.svelte
@@ -1,14 +1,15 @@
 <script lang="ts">
     import { twMerge } from "tailwind-merge";
 
-    let { class: className = "", children } = $props();
+    let { class: className = "", children, inlineStyles = "" } = $props();
 </script>
 
 <div
     class={twMerge(
-        "bg-light-surface border-light-muted min-w-sm rounded-[32px] border p-6 shadow-[0px_1px_3px_0px_rgba(0,0,0,0.1)]",
+        "bg-light-surface border-light-muted min-w-sm rounded-4xl border p-6 shadow-[0px_1px_3px_0px_rgba(0,0,0,0.1)]",
         className,
     )}
+    style={inlineStyles}
 >
     {@render children()}
 </div>
--- a/src/components/BaseCard.svelte
+++ b/src/components/BaseCard.svelte
@@ -1,15 +1,21 @@
 <script lang="ts">
-    import { twMerge } from "tailwind-merge";
+    import type { HTMLStyleAttributes } from "svelte/elements";
+    import { twMerge, type ClassNameValue } from "tailwind-merge";
 
-    let { class: className = "", children, inlineStyles = "" } = $props();
+    interface Props {
+        class?: ClassNameValue;
+        children: any;
+        style?: HTMLStyleAttributes["style"];
+    }
+    let { class: classes = "", children, style }: Props = $props();
 </script>
 
 <div
     class={twMerge(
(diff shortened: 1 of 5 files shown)


src/components/library/Category.svelte
```diff
@@ -4,7 +4,7 @@
     import type { Snippet } from "svelte";
 
     const styles = {
-        default: "inset-ring-1 hover:inset-ring-0 inset-ring-secondary bg-white",
+        default: "inset-ring-1 hover:inset-ring-0 inset-ring-secondary t bg-white text-secondary",
```
subiabre: What does the `t` class name do?

src/components/BaseCard.svelte
```diff
@@ -1,14 +1,15 @@
 <script lang="ts">
     import { twMerge } from "tailwind-merge";
 
-    let { class: className = "", children } = $props();
+    let { class: className = "", children, inlineStyles = "" } = $props();
```
subiabre: Other components in the code base try to name their props as close to native HTML tag attributes as possible. This component takes class name values in a prop externally named `class` and internally named `className` (though the convention seems to be on `classes`, but this is just me being nit-picky), why can't the inline styles be taken in a prop named `style`? At least externally.

While we are here, we could also add proper types to the props, `ClassNameValue` for `class` and `HTMLStyleAttributes` for `style`

src/pages/[...locale]/create/CreateProjectForm.svelte
```diff
@@ -321,26 +321,31 @@
...
     <div class="ml-auto">
-        <BaseCard class="flex h-full max-h-126.5 w-full max-w-109.25 flex-col">
+        <BaseCard
+            class="border-grey flex h-full max-h-126.5 w-full max-w-109.25 flex-col bg-white"
+            inlineStyles="box-shadow: 0 35px 10px 0 rgba(0, 0, 0, 0.00), 0 22px 9px 0 rgba(0, 0, 0, 0.01), 0 13px 8px 0 rgba(0, 0, 0, 0.05), 0 6px 6px 0 rgba(0, 0, 0, 0.09), 0 1px 3px 0 rgba(0, 0, 0, 0.10);"
+        >
             <h1
-                class="text-secondary text-2xl leading-8 font-bold {$currentDraft?.createProject
-                    .title || 'opacity-50'}"
+                class="text-secondary mb-2 text-[32px] leading-10 font-bold {$currentDraft
```
Comment (subiabre) [inline src/pages/[...locale]/create/CreateProjectForm.svelte]: `text-3xl` is valid tailwind and close enough to 32px

subiabre (changes requested):

src/components/library/Category.svelte
Dayv1dDev: t was being happy, that's its job xD

src/components/BaseCard.svelte
Dayv1dDev: Yep, this component was already created, I think are few more components which are using the className naming instead of classes, maybe we should create a task to simply change them all and continue with the standards. Now I'll modify the inline styles naming and giving the props types.

src/pages/[...locale]/create/CreateProjectForm.svelte
Comment (Dayv1dDev) [reply src/pages/[...locale]/create/CreateProjectForm.svelte]: I will add a new custom Tailwind class for the 32px common size used for texts that are titles or remarkable important information (as Tailwind do not bring us that option natively). This new custom TW class will be added in this PR: https://github.com/goteo/org.goteo.www/pull/282

subiabre (approved):

PR #288, opened 2026-05-18
Feat: add load more pagination project community

pepvillalonga:
---
  - https://app.asana.com/0/0/1214901282192618
  - https://app.asana.com/0/0/1215073905971249


src/components/project/ProjectCommunity.svelte
```diff
@@ -64,29 +63,43 @@
...
+        const items = (data as ProjectSupport[] | null) ?? [];
+        hasMore = items.length >= PAGE_SIZE;
+        return items;
+    }
 
+    async function enrichSupports(supports: ProjectSupport[]): Promise<EnrichedSupport[]> {
+        return Promise.all(
+            supports.map(async (support) => {
+                const id = extractId(support?.origin!);
                 const { data: user } = await apiUsersIdOrHandleGet({ path: { idOrHandle: id! } });
```
subiabre, line 68: The API documents the `origin` property of `ProjectSupports` as the Accounting where all the money in this support came from (https://api.v4.goteo.org/v4#tag/ProjectSupport/operation/api_project_supports_id_get). Then we are extracting the ID of the Accounting and using it to retrieve an User of that ID. This is broken, the user with ID "20" might own the accounting with ID "461". We need to check the owner of that Accounting and then show the name for it, when the owner is an User then we can use the `displayName` property, when the owner is a MatchCall we should use their `name`, etc.

Remember that if you need the API to return extra data under reasonable requests (such as this case, where showing a name would involve +1 extra call for each resource and then additional handling), then we can discuss how to better serve the necessary data from the API.

src/components/project/ProjectCommunity.svelte
```diff
@@ -64,29 +63,43 @@
...
+    async function fetchPage(page: number): Promise<ProjectSupport[]> {
+        const { data } = await apiProjectSupportsGetCollection({
+            query: { project: projectId, anonymous: false, page },
         });
 
-        const supportsWithOwners = await Promise.all(
-            (publicSupports || []).map(async (support) => {
-                const id = extractId(support?.origin!);
+        const items = (data as ProjectSupport[] | null) ?? [];
+        hasMore = items.length >= PAGE_SIZE;
```
subiabre: You don't need to recompute this every time. If you make the initial request with `application/ld+json` Content-Type header the API will return an enriched response body where the totality of items and pages will be given to you from the beginning.

In fact, this implementation is very weak and relies on the API having the value of `PAGE_SIZE` also as the default value of `itemsPerPage`, which might change per instance and even per-requests with the inclusion of a new value in the query param.

subiabre (changes requested):

src/components/project/ProjectCommunity.svelte
```diff
@@ -25,26 +28,23 @@
...
 
-    let projectsSupportItems = $state<
-        (ProjectSupport & {
-            displayName: string;
-            matchfunding: boolean;
-        })[]
-    >([]);
+    type EnrichedSupport = ProjectSupport & {
+        displayName: string;
+        matchfunding: boolean;
```
subiabre: Please regenerate the SDK and remove all the code you used to populate this data client side, now you can get both of these properties from the API.

subiabre (changes requested):

subiabre (approved):

PR #289, opened 2026-05-19
Feat/checkout login and register

angeldipinto:
The files login.astro and LoginForm.svelte were created, and register.astro and RegisterForma.svelte were modified.

And some others related to these four.

---
  - https://app.asana.com/0/0/1214301600008673


src/components/Checkout/Summary.svelte
```diff
@@ -8,13 +8,22 @@
...
-    );
+    const recipients = $derived(() => {
+        const data = { ...$cartByRecipient };
+
+        if (!data["foundation"]) {
+            data["foundation"] = [];
+        }
+        if (Object.keys(data).length === 1 && data["foundation"]) {
+            data["project_placeholder"] = [];
+        }
```
subiabre: What is this? Why are there hardcoded checks for a key `foundation`?

src/components/Checkout/Summary.svelte
```diff
@@ -43,21 +52,32 @@
...
-                                amount: `<strong>${formatCurrency(amount)}</strong>`,
-                                name: name,
-                            }}
-                        />
-                    </p>
-                {/each}
+            <div class="flex w-full flex-col pb-6">
+                <div class="bg-secondary mt-0 mb-4 h-px w-full max-w-[620px] border-none"></div>
+
+                <div class="text-content font-karla text-[16px] leading-6 font-normal">
```
subiabre: Do not hard-code these values anymore. Remember to **always** use the closest possible size spec in tailwind relative units. Also, do not refer explicitly to the font "Karla" anymore. Use the appropriate font name.

src/components/Checkout/Summary.svelte
```diff
@@ -43,21 +52,32 @@
...
+                        )}
+
+                        <span class="inline">
+                            <Thtml
+                                key={id === "foundation"
+                                    ? "checkout.summary.resume.foundationPrefix"
+                                    : "checkout.summary.resume.donationsPrefix"}
+                                vars={{
+                                    amount: `<strong class="font-bold text-secondary">${formatCurrency(amount)}</strong>`,
+                                }}
```
subiabre: I see where these changes come from now. The Figma design of this screen is outdated. Please, just use the current implementation of the Summary and remove the changes you've made that aren't strictly necessary for the login or register forms.

src/components/LoginForm.svelte
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    let { lang, currentStep = 2, session }: Props = $props();
+
+    let acceptTerms = $state(false);
+    let isSubmitting = $state(false);
+    let errorMessage = $state("");
+
+    onMount(() => {
+        localeStore.set(lang);
```
subiabre: The value of the locale store is already set on the Svelte context in the App.svelte component, which is pretty much always top-level and contains every other Svelte component. You should never re-set it unless you have a very good reason to do so. As is done in every other page, we start with a lean astro page that just contains the basic Layout, then inside it load the Svelte app: App.svelte component, and then the components necessary to compose the actual page. Check how the other pages in the /checkout are composed.

When in rome do as the romans do.

src/components/LoginForm.svelte
```diff
@@ -0,0 +1,167 @@
...
+    });
+
+    const handleSubmit = async (e: SubmitEvent) => {
+        e.preventDefault();
+        isSubmitting = true;
+    };
+
+    const inputClass =
+        "w-full rounded-xl border border-secondary bg-white px-4 py-4 text-base text-content placeholder:text-secondary/30 focus:ring-1 focus:ring-primary focus:outline-none transition-all shadow-sm";
```
subiabre: Why not just put these classes in the `class` attribute of the actual input tag?? Here it is just a magic string that tailwind extensions in the IDE can't understand, you lose the ability to know what each class does by hovering over them.

If your concern was that these classes get repeated for inputs that should look alike then: if the amount is low (2 or 3) copy pasting is fine. Alternatively you can just create an input component with the necessary styles: just like the `TextInput` component that we already have implemented in our library of components.

src/components/LoginForm.svelte
```diff
@@ -0,0 +1,167 @@
...
+        <div class="mx-auto max-w-353 px-6 lg:px-2">
+            <div class="flex flex-col gap-10 lg:flex-row lg:items-start">
+                <div class="flex flex-[0_0_1] flex-col items-start gap-10">
+                    <BackButton />
+
+                    <div class="flex flex-col items-start gap-10 self-stretch">
+                        <h1
+                            class="font-body self-stretch text-[40px] leading-12 font-bold text-[#3D3D3D]"
+                        >
+                            {$t("register.page.title")}
```
subiabre: Remember to use the `common`, `pages`, `domain`, `system` translation groups for your keys.

src/components/LoginForm.svelte
```diff
@@ -0,0 +1,167 @@
...
+                        <p
+                            class="text-content font-body self-stretch py-3 text-[14px] leading-4 font-medium"
+                        >
+                            {$t("checkout.summary.deduct")}
+                        </p>
+
+                        <Button
+                            type="submit"
+                            form="login"
+                            class="fflex bg-primary text-secondary font-karla items-center justify-center gap-2 self-start rounded-3xl px-6 py-4 text-[14px] leading-4 font-bold"
```
subiabre: `fflex` seems to be a typo of **flex**. I take the chance to remind you to check the classes you have defined and remove every explicit reference to theme values such as the Karla font and hardcoded, non-relative values such as sizes.

subiabre (changes requested):

PR #290, opened 2026-05-19
feat: layout anonymous donations card within community grid based on matchfunding presence

pepvillalonga:
---
  - https://app.asana.com/0/0/1214902077648324


src/components/project/ProjectCommunity.svelte
```diff
@@ -43,33 +43,24 @@
...
             },
             {} as Record<string, typeof projectsSupportItems>,
         ),
     );
 
-    async function fetchPage(page: number): Promise<ProjectSupport[]> {
+    async function fetchPage(page: number, size: number): Promise<ProjectSupport[]> {
         const { data } = await apiProjectSupportsGetCollection({
-            query: { project: projectId, anonymous: false, page },
+            query: { project: projectId, anonymous: false, page, itemsPerPage: size },
```
subiabre: 👍🏼 Much better than in the related PR, but still weak

src/components/project/ProjectCommunity.svelte
```diff
@@ -87,15 +78,28 @@
         );
     }
 
+    const matchfundingCount = $derived(groupedItems.matchfunding?.length ?? 0);
+    const hasMatchfunding = $derived(matchfundingCount > 0);
+    const isSingleMatchfunding = $derived(matchfundingCount === 1);
```
subiabre, line 76: Just like this solves nicely by hacing the API return all anonymous supports by allowing to filter by the `anonymous` property I feel like this would be better handled by having the API return the "matchfunding-ness" of ProjectSupports.

Not really in the scope of this task, but I'm drafting this PR and it shall be reworked once the API returns a better `ProjectSupport` shape.

subiabre (approved):

PR #301, opened 2026-05-25
Feat/unified search store query params api

angeldipinto:
The storesearch was unified to standardize API parameters, eliminating unnecessary translations.

---
  - https://app.asana.com/0/0/1214901203361630

### DIFF
--- a/src/stores/searchStore.ts
+++ b/src/stores/searchStore.ts
@@ -59,25 +59,25 @@ function loadInitialState(): SearchState {
     const urlParams = new URLSearchParams(window.location.search);
 
     // Initialize filters from URL parameters using API field names
-    state.filters.title = urlParams.get("title") || "";
-    state.filters.status = urlParams.get("status") || "";
+    if (state.filters) {
+        state.filters.title = urlParams.get("title") || "";
+        state.filters.status = urlParams.get("status") || "";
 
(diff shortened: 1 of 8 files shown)


src/stores/searchStore.ts
```diff
@@ -237,14 +238,16 @@ function createSearchStore() {
...
 
-            if (state.filters.query) {
-                urlParams.set("q", state.filters.query);
+            if (state.filters.title) {
+                urlParams.set("title", state.filters.title);
             }
-            if (state.filters.statusFilter) {
-                urlParams.set("status", state.filters.statusFilter);
+            if (state.filters.status) {
+                urlParams.set("status", state.filters.status);
```
subiabre: Doing "`if filter set("filter")` is not very scalable. We have a type (`SearchFilters = ApiProjectsGetCollectionData["query"]` which defines every single key and their possible types, let's use that type to build the url params. See [how to iterate over keys of objects](https://stackoverflow.com/questions/43389414/how-to-iterate-over-keys-of-a-generic-object-in-typescript) and then use the types to check on how should the params be appended based on their type in the store.

subiabre (changes requested):

subiabre (approved):

PR #303, opened 2026-05-27
Fix: update community card

pepvillalonga:
---
  - https://app.asana.com/0/0/1215073905971238


src/components/project/ProjectCommunity.svelte
```diff
@@ -64,28 +66,32 @@
...
+    $effect(() => {
+        apiProjectSupportsGetCollection({
             query: { project: projectId, anonymous: false },
+        }).then(async ({ data: publicSupports }) => {
+            const supportsWithOwners = await Promise.all(
+                (publicSupports || []).map(async (support) => {
+                    const id = extractId(support?.origin!);
+
+                    const { data: user } = await apiUsersIdOrHandleGet({
+                        path: { idOrHandle: id! },
```
subiabre: This is still mixing `Users` with `MatchCalls`, as both of them can be at the end of an `origin` in a `ProjectSupport`. If you need to enrich the item data with an avatar then it's probably to repeat the previous solution for `displayName` and `matchfunding` properties and have the API return a `displayImage`.

I'm drafting this PR until the API can also return that, then this will need to be re-thought to 1. stop mixing origin resources and 2. use the API-given data.

subiabre (changes requested):

src/components/project/ProjectCommunitySponsorModal.svelte
```diff
@@ -0,0 +1,42 @@
...
+    import { t } from "../../i18n/store";
+    import { formatCurrency } from "../../utils/currencies";
+    import Goteo from "../icons/Goteo.svelte";
+
+    import type { ProjectSupport } from "../../openapi/client/index";
+
+    let {
+        item,
+    }: {
+        item: ProjectSupport & { displayName: string; avatar: string | undefined };
```
subiabre: No longer necessary to extend the type, as the native type of `ProjectSupport` in the SDK already defines keys `displayName` and `displayImage`. Same error in the sponsor card, please remove it whole from all the places where this is done.

subiabre (changes requested):

subiabre (approved):

PR #306, opened 2026-06-01
Feat/(refactor) generic csv exporter & security fixes

angeldipinto:
It was modified to a generic and reusable component

---
  - https://app.asana.com/0/0/1215073905971230


src/components/admin/Contributions.svelte
```diff
@@ -72,39 +75,76 @@
...
+        baseUrl: "/api/relay",
+    });
 
     async function loadTotalTips() {
         if (!isEnabled || !$session) return;
-        const headers = $session.token.asHttpHeaders;
 
-        const { data: tipjar } = await apiTipjarsIdGet({ path: { id: tipjarId }, headers });
+        const { data: tipjar } = await apiTipjarsIdGet({
+            client: relayClient,
```
subiabre: You don't need to create a new client to switch the baseUrl, you can simply pass the API relay path as the argument of the `baseUrl` option to the API calls you wish to perform with it.

src/components/admin/Contributions.svelte
```diff
@@ -255,22 +290,17 @@
...
                 if (checkoutIri && !checkouts.has(checkoutIri)) {
-                    checkouts.set(checkoutIri, await fetchCheckout(checkoutIri, headers));
+                    checkouts.set(checkoutIri, await fetchCheckout(checkoutIri));
 
                     const originAccountingIri = checkouts.get(checkoutIri)?.origin;
 
                     if (originAccountingIri && !accountings.has(originAccountingIri)) {
-                        await preloadAccountingData(
-                            originAccountingIri,
-                            headers,
```
subiabre, line 295: Removing this header from the calls that use it breaks the entire /charges page because many lines of code rely on the extra metadata that the API returns when the request content is of `ld+json` type.

Please, always make sure you do not break the application while building on top of it. Check how the application behaves prior to your work and test thoroughly that none of the changes you have introduced create side-effects that you were not tasked with.

src/components/admin/ExportCsv.svelte
subiabre, line 1: This component is now specifically tied to `GatewayCharges` resources, but the scope of the task was to make it generic so that it could be easily plugged in on other admin pages and export the API contents exposed there.

Please, keep the working changes, but refactor the component so that it receives an operation endpoint (`/v4/charges`, `/v4/users`, etc), and a flexible `Record` type to pass query params. The component then fetchs the operation path with the query params via the /api/relay path. Finally make sure the transformation of data is flexible and recursively traverses nested JSON and destructures it into JSON-path names (example: `{ money: { amount: ... } }` -> `money.amount`, `{ territory: { country: ... } }` -> `territory.country`, etc).

subiabre (changes requested):

PR #308, opened 2026-06-02
Feat: implement layout admin projects

pepvillalonga:
---
  - https://app.asana.com/0/0/1215073905971249


src/components/Admin/FiltersTags.svelte
```diff
@@ -6,18 +6,19 @@
...
     import type { Locale } from "../../i18n/locales";
-    import type { ApiGatewayChargesGetCollectionData } from "../../openapi/client";
+
+    type Filters = Record<string, any>;
 
     let { title, filters, onCloseFilter } = $props<{
         title: string;
-        filters: ApiGatewayChargesGetCollectionData["query"];
-        onCloseFilter: (filters: ApiGatewayChargesGetCollectionData["query"]) => void;
+        filters: Filters;
```
subiabre, line 15: Actually, `ApiGatewayChargesGetCollectionData["query"]` is a much better typing. It contains every possible filter that the API takes as a key and it types each individual value for every key, and, as it comes from the SDK it will update automatically when the API changes. Please, always prefer the types from the SDK.

crowdin-test.yml
subiabre, line 1: I almost missed it! Please delete it.

subiabre (changes requested):

PR #310, opened 2026-06-03
feat: totalization tips

pepvillalonga:
---
  - https://app.asana.com/0/0/1215286452332643

### DIFF
--- a/src/components/Admin/Slider.svelte
+++ b/src/components/Admin/Slider.svelte
@@ -3,39 +3,32 @@
 
     import "flickity/css/flickity.css";
     import TotalizerCard from "./TotalizerCard.svelte";
+    import { session } from "../../auth/store";
     import { t } from "../../i18n/store";
+    import { apiAccountingsIdGet, apiTipjarsIdGet } from "../../openapi/client/sdk.gen";
     import { totalItems, isLoading } from "../../stores/chargesPaginationAndSort";
-    import { fetchAccounting, fetchTipjar } from "../../utils/cachedFetch";
     import { formatCurrency } from "../../utils/currencies";
-    import { isEnabled, tipjarIri } from "../../utils/tipping";
+    import { extractId } from "../../utils/extractId";
+    import { isEnabled, tipjarId } from "../../utils/tipping";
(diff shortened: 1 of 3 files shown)


src/components/Admin/Slider.svelte
```diff
@@ -5,13 +5,41 @@
...
+
+    const CACHE_NAME = "charges-cache";
 
     let mainCarousel: HTMLDivElement;
-    let flickity: Flickity;
     let isSliderLoaded = $state(false);
+    let totalTips = $state<string>("—");
+
+    function getAccessToken(): string | null {
+        const match = document.cookie.match(/(?:^|;\s*)access-token=([^;]*)/);
```
subiabre: This has several errors:

1. The cookie where the authentication data is stored is not `access-token`, it is `session`.
2. The `session` cookie where the authentication data is stored is http-only, so even if you looked for it with the correct name you would not be able to access it from svelte.
3. You can already obtain the structured, clean data from the `session` store, which is populated in the `App.svelte` component on almost every single page load from astro.
4. In the future, if the final goal of retrieving the access token is to perform an authenticated requests, it's better to simply switch the baseUrl of the operation call to the astro API relay: `/api/relay` (see src/pages/api/relay[...path].ts).

src/components/Admin/Slider.svelte
```diff
@@ -5,13 +5,41 @@
...
+        } catch {
+            return null;
+        }
+    }
+
+    async function loadTotalTips() {
+        if (!isEnabled) return;
+        const token = getAccessToken();
+        if (!token) return;
+        const tipjar = await fetchTipjar(tipjarIri, token, CACHE_NAME);
```
subiabre: These functionanilites are being deprecated by the work on https://github.com/goteo/org.goteo.www/pull/311 do not use these functions, simply use regular functions and if you want to use cache use the interceptor that you developed.

Good catch using the `tipjarIri` const from the tipping util.

subiabre (changes requested):

subiabre (approved):

PR #327, opened 2026-06-16
Fix: reset form fields on open

pepvillalonga:
---
  - https://app.asana.com/0/0/1215720778063729


src/components/project/edit/BudgetModal.svelte
```diff
@@ -35,6 +35,16 @@
 
     let openDeleteModal = $state(false);
 
+    $effect(() => {
+        if (open) {
+            selectedBudgetTitle = budgetItem?.title ?? "";
+            selectedBudgetType = budgetItem?.type;
+            amount = budgetItem?.money.amount ? budgetItem.money.amount / 100 : 0;
```
subiabre: Not every currency the app takes will be divisible by 100 smaller units, or even be divisible at all. Please use the appropriate function from the currency utils to format this amount and ensure that the amount sent to the API is expressed in the minor unit of the currency used.

subiabre (changes requested):

subiabre (approved):

PR #337, opened 2026-06-22
fix: replace deprecated apiUserTokensPost with OAuth password grant in register

pepvillalonga:
---
  - https://app.asana.com/0/0/1215799901595580

### DIFF
--- a/src/actions/register.ts
+++ b/src/actions/register.ts
@@ -2,6 +2,7 @@ import { z } from "astro/zod";
 import { ActionError, defineAction } from "astro:actions";
 
 import { passwordGrant } from "../auth/grant.ts";
+import { buildSession, setSession } from "../auth/session.ts";
 import {
     apiUsersPost,
     apiUsersIdpersonPatch,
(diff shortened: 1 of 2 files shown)


src/actions/register.ts
```diff
@@ -73,7 +66,7 @@ export const register = defineAction({
                 await apiUsersIdpersonPatch({
                     path: { id: userId },
                     headers: {
-                        Authorization: `Bearer ${authData?.token}`,
+                        Authorization: `Bearer ${auth.access_token}`,
```
subiabre: Because the `passwordGrant` function returns an `OAuthToken` object you can simply invoke the object's `asHttpHeaders` property which shall contain all the necessary headers to use the token in HTTP requests instead of having to manually wire the access token string into the correct shape and header.

src/actions/register.ts
```diff
@@ -105,21 +98,6 @@ export const register = defineAction({
                 });
             }
 
-            context.cookies.set(
```
subiabre, line 108: Deleting this is fine as it is old, but are you sure that the authentication that's been performed remains in session? Check `src/pages/[...locale]/login/callback.astro` which represents the final step in an *authorization code* grant and see how in there a session is built and set in cookies for the successful authentication flow. I don't see anything here persisting the authentication for the user.

subiabre (changes requested):

subiabre (approved):

PR #343, opened 2026-06-26
Chore: regroup project and auth keys into 4 categories

pepvillalonga:
---
  - https://app.asana.com/0/0/1213786273016337


src/components/Admin/Search.svelte
```diff
@@ -200,8 +200,8 @@
                                         )}
                                     </div>
                                     <div class="mt-1 text-sm text-gray-500 italic">
-                                        {$t("domain.charges.entityLabels.tipjar-id")}: {item
-                                            .data.id}
+                                        {$t("domain.charges.entityLabels.tipjar-id")}: {item.data
```
subiabre: Don't forget the camelCase convention!

src/components/Admin/Slider.svelte
```diff
@@ -54,12 +54,12 @@
...
-                title: $t("admin.projects.totalizers.selected"),
+                title: $t("pages.admin.projects.totalizers.selected"),
                 amount: $totalItems,
             },
-            { title: $t("admin.charges.totalizers.totalCharges"), amount: "—" },
-            { title: $t("admin.charges.totalizers.totalTips"), amount: totalTips },
-            { title: $t("admin.charges.totalizers.totalFees"), amount: "—" },
+            { title: $t("domain.charges.totalizers.totalCharges"), amount: "—" },
+            { title: $t("domain.charges.totalizers.totalTips"), amount: totalTips },
+            { title: $t("domain.charges.totalizers.totalFees"), amount: "—" },
```
subiabre: I totally think this should not be in domain as it does not represent anything of the logic model of the charges itself, it rather is a pages key, as it shows as the title for one of the totalizers boxes in the admin page. Please, update similar keys with the same domain/pages mix problem.

src/components/library/Search.svelte
```diff
@@ -22,7 +22,7 @@
     let {
         class: classes = "",
         value = $bindable(""),
-        placeholder = $t("search.placeholder"),
+        placeholder = $t("common.navigation.search.placeholder"),
```
subiabre: I do think navigation is domain, while definitely common because it is all across the app like generic action/help messages, it also represents a core logical piece of the application: it's the map and compass that will guide most users browse across the app.

crowdin.yml
```diff
@@ -1,3 +1,3 @@
 files:
-  - source: /src/i18n/locales/es.json
-    translation: /src/i18n/locales/%two_letters_code%.json
+    - source: /src/i18n/locales/es.json
+      translation: /src/i18n/locales/%two_letters_code%.json
```
subiabre, line 3: Will this not be an issue with prettier?

subiabre (changes requested):

PR #347, opened 2026-07-01
feat: implement new datainput

pepvillalonga:
---
  - https://app.asana.com/0/0/1216151185966162

### DIFF
--- a/src/components/library/inputs/DateInput.svelte
+++ b/src/components/library/inputs/DateInput.svelte
@@ -1,8 +1,9 @@
 <script lang="ts">
-    import { Datepicker } from "flowbite-svelte";
+    import { clickOutside, Datepicker } from "flowbite-svelte";
     import { twMerge, type ClassNameValue } from "tailwind-merge";
 
     import { locale, t } from "../../../i18n/store";
+    import { formatDate } from "../../../utils/dates";
     import Calendar from "../../icons/Calendar.svelte";
     import Close from "../../icons/navigation/Close.svelte";
 
@@ -50,7 +51,7 @@
     let container: HTMLDivElement;
 
     const displayValue = $derived(
-        value && !isNaN(value.getTime()) ? value.toLocaleDateString($locale) : "",
+        value && !isNaN(value.getTime()) ? formatDate(value, $locale) : "",
     );
 
     function dateToString(date: Date): string {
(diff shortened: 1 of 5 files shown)


src/components/library/inputs/DateInput.svelte
```diff
@@ -47,86 +63,190 @@
         return `${year}-${month}-${day}`;
```
subiabre, line 64: I know this is old already, but I am commenting on this one so that it does not get accidentally lost.

Could you update the date formatting code across this PR? Simply replace it by the `formatDate` from "utils/dates.ts", the goal is to have less duplicate code and dates being formatted in a plethora of different ways for no reason across the app.

src/components/library/inputs/DateInput.svelte
```diff
@@ -47,86 +63,190 @@
...
-            }
-            return;
+    function apply() {
+        onApply?.(value);
+        onBlur?.();
+        open = false;
+    }
+
+    function calendarMonth(root: HTMLElement): { year: number; month: number } | null {
+        const text = root.querySelector('[aria-live="polite"]')?.textContent?.toLowerCase();
```
subiabre: I assume you are doing this DOM manipulation instead of simply binding to the actual element because the inner date picker component does not provide a way of binding to that element. But here `root` is binded to a `div` element which contains many other things and I am concerned on how safe is this selection. Are you sure doing DOM manipulation is the only way? If that is the case can you improve the querySelector expression to narrow it further? Something like the element ID as its done in the `applyLimit` function.

src/components/library/inputs/DateInput.svelte
```diff
@@ -47,86 +63,190 @@
...
-    // Convert min/max to string format for HTML input
-    const minString = $derived(toDateString(min));
-    const maxString = $derived(toDateString(max));
+    $effect(() => {
+        if (!open) return;
+        const onPointerDown = (event: PointerEvent) => {
+            if (container && !container.contains(event.target as Node)) open = false;
+        };
+        document.addEventListener("pointerdown", onPointerDown);
+        return () => document.removeEventListener("pointerdown", onPointerDown);
```
subiabre: Not a fan of globally attaching event listeners to the DOM. Can you use the `clickOutside` svelte directive? (as it's done in many Dropdown components) If not I would attach if possible this event listener to the `root` element and prevent race conditions bugs with an early return of this effect when the the DOM is not yet ready (Astro could attempt to perform something on this component in SSR and it would be a very annoying bug).

subiabre (changes requested):

src/components/library/inputs/DateInput.svelte
```diff
@@ -91,8 +92,14 @@
         open = false;
     }
 
+    // The month label is the <h3 aria-live="polite"> that flowbite-svelte renders
+    // inside the navigation row of #datepicker-dropdown.
+    function monthLabel(root: HTMLElement): HTMLElement | null {
+        return root.querySelector(':scope > div > h3[aria-live="polite"]');
```
subiabre: Much better now.

However, and sorry for not realizing this myself earlier, but could you also throw an exception or log an error when this functions fails to find the element it needs? We are generating a dependency in here and betting that the inner date picket HTML will always remain compatible with our selection query, but that can change and we would hardly notice it if we go and run `pnpm update`.

subiabre (changes requested):

subiabre (approved):

PR #350, opened 2026-07-02
fix: update-project-states-on-search-filters

pepvillalonga:
---
  - https://app.asana.com/0/0/1215553690534665


src/components/library/dropdown/DropdownMenu.svelte
```diff
@@ -47,6 +51,10 @@
             }
         }
 
+        if (sortSelectedFirst) {
+            merged.sort((a, b) => Number(selectedIds.has(b.id)) - Number(selectedIds.has(a.id)));
```
subiabre: You could skip this sorting and avoid weird bugs caused by the comparison expression by simply flipping the building logic around. If the `selectedFirst` prop is `true` then the `merged` array is built with the selected items first, which should be faster than pushing them to the end of the array and then sorting them to first place.

src/utils/projectStatus.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * Public project states for the public search page.
+ *
+ * A project has many internal `status` values across its lifecycle. Only three
+ * grouped "states" are exposed to public search. The UI multiselect works with
+ * these states; the API receives the expanded `status[]` list.
+ */
+export type PublicProjectState = "in_campaign" | "funded" | "archived";
```
subiabre: This entire PR mixes "state" and "status" almost interchangeably. While I understand how using "state" for the publicly shown project statuses might help avoiding confusion I'm also afraid that this could very well turn into a very subtle bug that is a nightmare to diagnose.

May I suggest a better naming: **public status**, so: `type ProjectPublicStatus`, `const PUBLIC_STATUSES_TO_STATUSES`, `const ALL_PUBLIC_STATUSES`, etc.

src/stores/searchStore.ts
```diff
@@ -62,6 +76,9 @@ export function parseSearchParamsFilters(searchParams: URLSearchParams) {
         }
     }
 
+    // Public search always constrains status to the union of the selected states.
+    filters["status[]"] = expandStatesToStatuses(parseStatesFromParams(searchParams));
```
subiabre: This store is already large enough as it is, and being honest there is only one place when the public statuses feature will be needed: the search page.

Please, keep the _public states_ -> _API project status_ mapping outside of this store. You can replicate the same strategy used by the TerritoryFilter, where all the logic to go from plain text addresses to the ISO 3166 codes that the API understands happens inside the component and this one simply gives back the values that the store expects.

subiabre (changes requested):

src/services/projectsService.ts
```diff
@@ -29,6 +30,9 @@ export class ProjectsService {
             const response = await apiProjectsGetCollection({
                 query: {
                     ...filters,
+                    // Public search never queries a non-public status, regardless of
+                    // what the filters (e.g. from the URL) contain.
+                    "status[]": constrainToPublicStatuses(filters?.["status[]"]),
```
subiabre: There is also the `status` (without the brackets `[]`) query param. You can see all available query params in the API docs https://api.v4.goteo.org/v4#tag/Project/operation/api_projects_get_collection (also in the SDK which is also the basis for the `SearchFilters` type used in the search store and this service.

src/stores/searchStore.ts
```diff
@@ -512,16 +472,14 @@ export const isEmpty = derived(
...
-    const { filters, selectedStates } = $searchStore;
-    // States count as an active filter only when a strict, non-empty subset is chosen.
-    const statesActive =
-        selectedStates.length > 0 && selectedStates.length < ALL_PUBLIC_STATES.length;
-    return !!(filters?.title || statesActive || (filters?.["categories[]"]?.length ?? 0) > 0);
+    const { filters } = $searchStore;
+    return !!(
+        filters?.title ||
+        (filters?.["status[]"]?.length ?? 0) > 0 ||
+        (filters?.["categories[]"]?.length ?? 0) > 0
```
subiabre: This is weak (there are other available filters which could have a value and this function would still say there are no active filters) and does not scale (every time a new filter is added to the Projects collection endpoint in the API a new line has to be added).

One of the solutions explained in https://builtin.com/software-engineering-perspectives/javascript-check-if-object-is-empty should provide a better way to do this.

subiabre (changes requested):

subiabre (approved):

PR #354, opened 2026-07-03
[fix] bypass browser cache interceptor on direct read operations in admin projects page

Dayv1dDev:
## Summary

The browser cache interceptor was applied globally to all GET requests, including direct read operations where users expect fresh data (search by name, expanding project cards). This PR extracts the interceptor lifecycle into a shared module and adds a `withoutCache()` helper to bypass the cache for specific direct-read operations on the admin projects page.

## Behaviour

| Operation | Cache |
|---|---|
| General project list load | Cached |
| Search by project name | Bypassed |
| Expand project row (owner fetch) | Bypassed |

---
  - https://app.asana.com/0/0/1215921687904493

### DIFF
--- a/src/components/admin/Projects.svelte
+++ b/src/components/admin/Projects.svelte
@@ -5,6 +5,7 @@
     import ProjectsTable from "./ProjectsTable.svelte";
     import Slider from "./Slider.svelte";
     import { t } from "../../i18n/store";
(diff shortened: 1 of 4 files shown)


subiabre (approved):
Very elegant solution! 👍🏼

PR #355, opened 2026-07-03
[feat] Implement dinero.js + replace arithmetic operations and avoid .amount usage

Dayv1dDev:
---
  - https://app.asana.com/0/0/1216074825280345


src/utils/money.ts
```diff
@@ -0,0 +1,91 @@
...
+function getCurrency(code: string): DineroCurrency {
+    if (!(code in currencies)) {
+        throw new Error(`Unknown currency code: ${code}`);
+    }
+    return currencies[code as keyof typeof currencies] as DineroCurrency;
+}
+
+function toDinero(money: Money) {
+    const amount = money.amount ?? 0;
+    const currency = getCurrency(money.currency ?? "EUR");
```
subiabre: Do not hardcode values, ever. If you need a fallback value use the appropriate env var.

src/utils/money.ts
```diff
@@ -0,0 +1,91 @@
...
+export function subtractMoney(a: Money, b: Money): Money {
+    return toMoney(subtract(toDinero(a), toDinero(b)));
+}
+
+export function multiplyMoney(m: Money, factor: number): Money {
+    return toMoney(multiply(toDinero(m), factor));
+}
+
+export function sumMoney(items: Money[]): Money {
+    if (items.length === 0) return { amount: 0, currency: "EUR" };
```
subiabre: Again, do not hardcode values, ever.

src/components/checkout/PaymentMethod.astro
```diff
@@ -170,9 +170,10 @@ const { accounting } = Astro.props;
         if (cartData && walletInput) {
             try {
                 const cart = JSON.parse(cartData);
-                const total = cart.items.reduce(
-                    (sum: number, item: { amount: number; quantity: number }) =>
-                        sum + item.amount * item.quantity,
+                const items = Object.values(cart.items ?? {});
+                const total = items.reduce(
+                    (sum: number, item: { money: { amount: number }; quantity: number }) =>
+                        sum + item.money.amount * item.quantity,
```
subiabre: Shouldn't this use the `sumMoney` and `multiplyMoney` functions?

subiabre (changes requested):
Please, fix the comments and also review the code and remove all hardcoded instances of "EUR" before I finish reviewing this PR.

src/components/checkout/PaymentMethod.astro
Dayv1dDev: That's right, my fault

src/utils/money.ts
Dayv1dDev: All the hardcoded "EUR" values are for a reason, I asked you in previous tasks about currencies and you answered that Euro is our standard currency if no one is specified. But as you said now, I think it's better to get it from env vars, as the code core works for other future/recent platforms. Just to point that I was aware about the hardcoding thing.

subiabre (approved):

PR #358, opened 2026-07-08
[feat] Images management in project form

Dayv1dDev:
---
  - https://app.asana.com/0/0/1215720778063731


src/utils/imageUpload.ts
```diff
@@ -36,28 +37,46 @@ export async function uploadImage(file: File, options?: UploadOptions): Promise<
 
     options?.onProgress?.("uploading");
 
-    const putRes = await fetch(signedUrl, {
-        method: "PUT",
-        headers: { "Content-Type": file.type },
-        body: file,
-    });
+    const { url, key } = await new Promise<UploadResult>((resolve, reject) => {
+        const xhr = new XMLHttpRequest();
```
subiabre, line 47: Why XHR instead of fetch?

subiabre (changes requested):
Implementation is also missing showing the uploaded cover image of rewards as designed in Figma.

src/utils/imageUpload.ts
Dayv1dDev: I'm using XML instead of fetch because fetch does not expose upload progress events. This callback is consumed by `FileUpload.svelte` to render a real-time progress bar with percentage. I left a comment on the code clarifying the XMLHttpRequest purpose.

subiabre (approved):

PR #360, opened 2026-07-08
[fix] Show owner display name filter tag + fix pagination handling with JSON-LD in admin/charges

Dayv1dDev:
---
  - https://app.asana.com/0/0/1216299428597489


src/components/admin/Contributions.svelte
```diff
@@ -99,8 +99,14 @@
         return query;
     }
 
-    function getCollectionTotalItems(collection: unknown): number {
-        if (Array.isArray(collection)) return collection.length;
+    function getCollectionTotalItems(collection: unknown, response?: Response): number {
+        if (Array.isArray(collection)) {
+            const headerTotal =
+                response?.headers.get("X-Total-Items") ??
+                response?.headers.get("Content-Range")?.split("/")[1];
```
subiabre: The `X-Total-Items` headers is not in the response of any API response, and the `Content-Range` header will never be returned (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Range)

src/components/admin/Contributions.svelte
```diff
@@ -214,9 +220,18 @@
...
 
             const query = buildChargesQuery(filters, page, items);
-            const headers = $session?.token.asHttpHeaders as HeadersInit | undefined;
-
-            const { data: collection, error } = await apiGatewayChargesGetCollection({
+            const authHeaders = $session?.token.asHttpHeaders as Record<string, string> | undefined;
+
+            const headers = {
+                Accept: "application/ld+json",
+                ...(authHeaders ?? {}),
```
subiabre: Why declare the `authHeaders` const if only used here for destructuring? Why loosen the type of it as a generic Record instead of leaving the actual type from the `OAuthToken` type? Could you not have simply done:

```ts
apiGatewayChargesGetCollection({
    headers: {
        ...$session?.token.asHttpHeaders,
        Accept: "application/ld+json"
    }
})
```

src/components/admin/Contributions.svelte
```diff
@@ -364,6 +380,16 @@
...
 
+    let prevItemsPerPage = $state($itemsPerPage);
+
+    $effect(() => {
+        const current = $itemsPerPage;
+        if (current !== prevItemsPerPage) {
+            prevItemsPerPage = current;
+            $currentPage = 1;
+        }
+    });
```
subiabre, line 386: The lines above could have potential merge conflicts with https://github.com/goteo/org.goteo.www/pull/356, which is already approved and pending merge. Please, merge the working branch of that PR into this branch to avoid issues.

subiabre (changes requested):

src/components/admin/Contributions.svelte
Dayv1dDev: I've done the merge and there are no merge conflicts existing :+1:

src/components/admin/Contributions.svelte
Dayv1dDev: `headers` const is being used through all `loadCharges` function (multiple times) and it makes this large function code more legible. That's why I think we should mantain at least `headers` const (removing `authHeaders`).

subiabre: My bad, I actually checked for the wrong var when reviewing the code in my IDE.

subiabre (approved):

PR #363, opened 2026-07-10
feat: add output estimate in step form

pepvillalonga:
---
  - https://app.asana.com/0/0/1216207405325071

### DIFF
--- a/src/components/project/edit/ConfigurationStep.svelte
+++ b/src/components/project/edit/ConfigurationStep.svelte
@@ -4,6 +4,7 @@
     First step of the project setup wizard.
     Handles:
     - Categories (up to 2)
+    - Campaign release date
     - Funding rounds (1 or 2)
 
     Validation:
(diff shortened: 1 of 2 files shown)


src/components/project/edit/ConfigurationStep.svelte
```diff
@@ -112,6 +135,24 @@
...
+    <div class="space-y-4">
+        <div class="space-y-4">
+            <h2 class="text-2xl font-bold text-black">
+                {$t("pages.project.create.release.title")}
+            </h2>
+            <p class="text-content text-base font-normal">
+                {$t("pages.project.create.release.subtitle")}
+            </p>
+        </div>
+        <DateInput
```
subiabre, line 148: The DateInput component grows as large as the container, as it should. But notice how the step 2 limits the width to `max-w-167`, please do the same here.

subiabre (changes requested):

subiabre (approved):

PR #364, opened 2026-07-10
fix(project form): project edit form issues

subiabre:
---
  - https://app.asana.com/0/0/1216436149901509

### DIFF
--- a/src/components/project/edit/BudgetModal.svelte
+++ b/src/components/project/edit/BudgetModal.svelte
@@ -138,6 +138,7 @@
         <TextInput
             bind:value={amount}
             type="number"
+            labelText={$t("pages.project.edit.budget.modal.placeholders.moneyAmount")}
             placeholder={$t("pages.project.edit.budget.modal.placeholders.moneyAmount")}
             error={amountError}
         />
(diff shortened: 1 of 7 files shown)


src/components/project/edit/BudgetModal.svelte
```diff
@@ -149,54 +121,34 @@
...
-                    </option>
-                {/if}
-                <option value="minimum">
-                    {$t("domain.project.budget.minimum")}
-                </option>
-                <option value="optimum">
-                    {$t("domain.project.budget.optimum")}
-                </option>
-            </select>
-        </div>
```
Dayv1dDev, line 192: This part from `BudgetModal` is deleted on purpose? Are we giving up that fields? (Just to know)

subiabre: @Dayv1dDev yes, it's deleted on purpose, but not fully removed. The options selector was opaque to users as it was disabled and did not show any of the deadline options. Tests with users showed that they could not understand the input was disabled because they already had indirectly selected "minimum" or "optimum" based on which creation card they chose. The data is instead supplied by the modal based on the defaults that are set on the creation card itself.

Dayv1dDev (approved):

PR #365, opened 2026-07-13
[feat] Use Nominatim to convert addresses in plain text to standardized adresses + to API

Dayv1dDev:
---
  - https://app.asana.com/0/0/1216207405325073


src/components/library/inputs/AddressAutocomplete.svelte
```diff
@@ -0,0 +1,156 @@
...
+        name?: string;
+        error?: string;
+        onAddressChange?: (address: string, territory: Territory) => void;
+        onBlur?: () => void;
+    }
+
+    let {
+        class: classes = undefined,
+        value = $bindable(""),
+        placeholder = "Busca una dirección...",
```
subiabre: Remember to avoid hardcoded values, even for fallbacks and defaults. If you really wish to set a default value, use a translation key to supply the default value here, otherwise you are forcing every user of this component to remember to supply a translated placeholder every single time.

src/components/library/inputs/AddressAutocomplete.svelte
```diff
@@ -0,0 +1,156 @@
...
+                    </button>
+                {/each}
+            </div>
+        </div>
+    {/if}
+
+    {#if error}
+        <p class="mt-1 ml-4 text-xs text-red-600" role="alert">{error}</p>
+    {/if}
+</div>
```
subiabre, line 101: Why not build this component on top of the exisiting `DropdownMenu` component?? It gives better DOM reactivity, is closer to the component design in Figma and it would make it easier to keep consistent styling across the app, all of it while writing less code.

If there is something from that component that you are missing here, as I think it's the case with the close button, at least try to use other existing components instead of raw HTML form tags, like `Search` and `Button` (for the same consistency advantage), or even better: update the `DropdownMenu` to include or allow you to include that close button.

Re-implementing inputs over and over across the app makes maintenance and debugging harder, creates smelly patterns like `INPUT_CLASSES` constants to store the tailwind classes which can´t be easily debugged or understood by static analysis tools, and increases the cost of brand-tailoring tasks in an application that can, in theory, be shipped under other owners. On top of making the effort of developing the components library feel like a waste of time.

src/components/project/edit/OwnerInfoStep.svelte
```diff
@@ -175,28 +165,21 @@
             </div>
         </div>
 
-        <!-- Lugar de actividad
+        <!-- Lugar de actividad -->
```
subiabre: In english, please. Source code is always in english. This application is not only open-source and shall be understood by as much people as possible, it's an open-core that is already awaited by partners who do not speak spanish.

src/components/library/inputs/AddressAutocomplete.svelte
```diff
@@ -0,0 +1,156 @@
...
+    let lastSelectedAddress = $state("");
+
+    function extractTerritory(result: NominatimResult): Territory {
+        const address = result.address ?? {};
+        const country = address.country_code?.toUpperCase() ?? null;
+
+        const iso3166_2 = Object.entries(address)
+            .filter(([key]) => key.startsWith("ISO3166-2-"))
+            .sort()
+            .map(([, v]) => v);
```
subiabre: Please, check the code in TerritoryFilter.svelte, which does exactly the same. It feels better to have this outsourced to the nominatim service and have both components use the same process to extract ISO3166 codes out of nominatim results.

src/i18n/locales/es.json
```diff
@@ -626,6 +627,11 @@
           "title": "¿Qué categoría se ajusta a tu campaña?",
           "subtitle": "(Puedes elegir un máximo de 2)"
         },
+        "address": {
+          "title": "¿Cuál es la dirección del proyecto?",
+          "subtitle": "Busca y selecciona la dirección completa del proyecto. Esto nos ayuda a geolocalizar tu iniciativa.",
```
subiabre: Please avoid geolocation references
> Esto nos ayuda a ~~geo~~localizar tu iniciativa

...or something other avoiding any [geoposition](https://en.wikipedia.org/wiki/Geopositioning) synonym. It can be a sensitive topic and we don't really perform any related technique to extract this information from our users, we only rely on the info they supply on their own.

subiabre (changes requested):

src/components/project/edit/OwnerInfoStep.svelte
Dayv1dDev: That comment and others were done by another dev when creating OwnerInfoStep component, I only passed prettier to the changes I've done to that file. Anyway I should have seen that and remove it/change it.

src/components/library/inputs/AddressAutocomplete.svelte
Dayv1dDev: Didn't meant to do that

src/i18n/locales/es.json
Dayv1dDev: "Ubicar" or "situar" could be better, right?

subiabre (approved):

PR #366, opened 2026-07-14
fix: correct copy and count for selected projects card in admin charges

pepvillalonga:
---
  - https://app.asana.com/0/0/1212998936497211


subiabre (changes requested):
Please, update the SDK to the shape of https://github.com/goteo/org.goteo.api/pull/218 and remove all the fallback code: it increases the complexity of the already large Contributions.svelte for very little gain as we already have a stable method to get the projects count from the API.

subiabre (approved):

PR #367, opened 2026-07-14
[feat] {Composed Filters} Implement filterComposer util with subjects registry + helper functions + interfaces/types

Dayv1dDev:
I've updated the docs too with the changes and decisions made during implementation. Here is the reason to use `string` type instead of `keyof ApiProjectsGetCollectionData["query"]` in `FilterSubject.key`: https://docs.platoniq.net/doc/composed-filters-lwDHlkg9ci/#h-filtersubjectkey-string-vs-keyof-apiprojectsgetcollectiondataquery

---
  - https://app.asana.com/0/0/1216505761591940


src/utils/filterComposer.ts
```diff
@@ -0,0 +1,146 @@
+export interface FilterRow {
+    id: string;
+    subject: FilterSubject;
+    operator: FilterOperator;
+    referent: string | string[];
```
subiabre: Why not allow for numbers and dates (as the subjects already define those types) and then let the serialization take care of transforming those types to strings?

src/utils/filterComposer.ts
```diff
@@ -0,0 +1,146 @@
...
+        key: "dateUpdated",
+        type: "date",
+        compatibleOperators: ["before", "after", "strictly_before", "strictly_after"],
+    },
+    budgetAmount: {
+        key: "budgetAmount",
+        type: "number",
+        compatibleOperators: ["gte", "gt", "lte", "lt"],
+    },
+};
```
subiabre, line 214: I'm missing a way of telling these Project-related filter subjects apart from subjects of other resources, such as those that were already documented for the GatewayCharge filtering.

src/utils/filterComposer.ts
```diff
@@ -0,0 +1,146 @@
...
+            switch (operator) {
+                case "equals":
+                    return { [key]: referent as string };
+                case "is_any_of":
+                    return { [`${key}[]`]: referent as string[] };
+                case "before":
+                case "after":
+                case "strictly_before":
+                case "strictly_after": {
+                    const apiKey = key === "releaseDate" ? "dateCreated" : key;
```
subiabre: Not a fan of having this var named `apiKey`, that name is usually reserved for credentials to use an API, the API itself refers to these as query parameters, not keys.

Also, `releaseDate` does not really map to `dateCreated` (IMPORTANT the `date*` fields in the API are not set externally, they are timestamps set by the API itself, for all resources). It actually maps to `calendar.release`, which I have noticed is not present as a filter in the API.

subiabre (changes requested):

subiabre (approved):

PR #369, opened 2026-07-16
feat: add configurable campaign end date limit

pepvillalonga:
---
  - https://app.asana.com/0/0/1212073861451376

### DIFF
--- a/src/pages/[...locale]/create/validation.ts
+++ b/src/pages/[...locale]/create/validation.ts
@@ -1,6 +1,6 @@
 import { z } from "zod";
 
-import { getMaxReleaseDate } from "../../../utils/campaign";
+import { maxEndDate } from "../../../utils/campaign";
 
 /**
  * Validation schema for project creation form.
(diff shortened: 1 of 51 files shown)


src/pages/[...locale]/create/validation.ts
```diff
@@ -58,6 +60,21 @@ export const projectCreationSchema = z.object({
             },
             { message: "system.validation.project.release.min" },
         )
+        .refine(
+            (date) => {
```
Comment (subiabre) [inline src/pages/[...locale]/create/validation.ts:73]: Does this validation only take place when the max date is enabled?

src/utils/campaign.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * Campaign duration in days for each funding round choice.
+ * Mirrors the server-side calendar computation: the minimum deadline is
+ * 40 days after release, the optimum deadline 40 days after the minimum.
+ */
+export const CAMPAIGN_DURATION_DAYS: Record<"minimum" | "optimum", number> = {
+    minimum: 40,
+    optimum: 80,
```
subiabre: These are defaults based on the platform experience, but they could change and be shorter or longer for many projects. The only goal really is to avoid having campaigns across 2 different calendar years, we don't really want to enforce such strict campaign deadlines.

subiabre (changes requested):

src/pages/[...locale]/create/validation.ts
Comment (pepvillalonga) [reply src/pages/[...locale]/create/validation.ts:73]: Yes, it only runs when a cutoff is configured. After refactoring this to only use the max date, the guard is now if (!maxEndDate) return true, so when PUBLIC_CAMPAIGN_MAX_END_DATE is empty the check is skipped entirely.

subiabre (approved):

PR #372, opened 2026-07-16
fix: crowdin resolve errors

pepvillalonga:
---
  - https://app.asana.com/0/0/1216591872203787

### DIFF
--- a/.github/workflows/crowdin-upload.yml
+++ b/.github/workflows/crowdin-upload.yml
@@ -25,7 +25,7 @@ jobs:
                   download_translations: false
                   push_translations: false
                   create_pull_request: false
-                  crowdin_branch_name: develop
+                  crowdin_branch_name: ${{ github.ref_name }}
               env:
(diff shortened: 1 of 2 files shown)


.github/workflows/crowdin-upload.yml
```diff
@@ -25,28 +25,34 @@ jobs:
                   download_translations: false
                   push_translations: false
                   create_pull_request: false
-                  crowdin_branch_name: develop
+                  crowdin_branch_name: ${{ github.ref_name }}
```
subiabre, line 28: Wouldn't this trigger new translation PRs on every change across all working branches?

pepvillalonga: No, the push trigger is restricted to develop, so github.ref_name is always develop on automatic runs. Other branches only run via manual workflow_dispatch. Also this workflow has push_translations: false and create_pull_request: false; it just uploads the source file and kicks off AI pre-translation, so it never opens translation PRs. The change keeps the branch name consistent with the downstream CROWDIN_BRANCH steps when dispatched manually.

PR #392, opened 2026-07-30
Feat: tiptap rich text integration

pepvillalonga:
---
  - https://app.asana.com/0/0/1213999847221738


src/i18n/locales/es.json
```diff
@@ -138,6 +138,16 @@
       "nextMonth": "Mes siguiente",
       "close": "Cerrar"
     },
+    "richTextEditor": {
```
subiabre, line 133: "rich text editor" is 100% a product-domain thing rather than a commonly seen collection of messages.

src/components/project/edit/RichTextEditor.svelte
```diff
@@ -87,294 +62,240 @@
...
 
-<div class={twMerge("rich-text-editor space-y-4", className)}>
-    <!-- Toolbar -->
+{#snippet toolbarButton({ labelKey, active, run, align, glyph }: ToolbarButton)}
+    <button
+        type="button"
+        onclick={run}
+        class={twMerge(
+            "flex size-10 cursor-pointer items-center justify-center rounded-lg border bg-white shadow-sm",
+            active ? "border-secondary" : "border-grey",
```
subiabre, line 207: For cases like these, where there is not a possible class conflict you can use the `twJoin` function (https://github.com/dcastil/tailwind-merge/blob/main/docs/when-and-how-to-use-it.md#joining-internal-classes), which will compute the final set of classes but is much more performant.

Please, replace similar ocurrences in this PR.

subiabre (changes requested):

src/i18n/locales/es.json
```diff
@@ -147,6 +138,15 @@
       "alignCenter": "Centrar",
       "alignRight": "Alinear a la derecha",
       "characterCount": "{{ current }} / {{ max }} caracteres"
+    }
+  },
+  "common": {
+    "_": "Traducciones globales de la interfaz (botones, etiquetas, etc.)",
+    "dateInput": {
```
subiabre, line 150: A "date input" group for our specific DateInput component belongs to domain, just like "richTextEditor". Yes, they are common, but not because they are commonly seen call to actions or labels, they are common because they are part of specific, product-domain, features of our application.

subiabre (changes requested):

subiabre (approved):

PR #394, opened 2026-08-04
Chore: eslint unused definitions and svelte chacks

pepvillalonga:
---
  - https://app.asana.com/0/0/1216951879214119


src/components/admin/Users.svelte
```diff
@@ -74,7 +74,9 @@
 
         const sortOption = sortMap[sort];
         if (sortOption) {
-            query[`order[${sortOption.field}]`] = sortOption.direction;
+            // The spec only declares `order[handle]`, so the email variant is written
+            // through a record view of the query.
+            (query as Record<string, unknown>)[`order[${sortOption.field}]`] = sortOption.direction;
```
subiabre: There are no plans to support ordering by email, querying by email is delicate enough and we had to do a set of PRs in the API in order to prevent 3rd parties from ex-filtrating email addresses in any way. We should remove it rather than give it support.

src/components/project/edit/ProjectEditor.svelte
```diff
@@ -69,6 +68,27 @@
...
 
+    /**
+     * The cover comes back as a bare URL, but `UploadedFile` needs a MIME type
+     * for the uploader to recognise it as an image and render its preview.
+     */
+    function guessImageMimeType(url: string): string {
+        const extension = url.split("?")[0].split(".").pop()?.toLowerCase();
+
+        switch (extension) {
+            case "png":
```
subiabre: This entire block feels weak. Please consider this answer for extracting file extension from paths: https://stackoverflow.com/a/12900504

Also the switch statement is not synced with the allowed types declared in src/utils/objectStorage.ts, this function is guessing types that will probably never have to deal with.

src/components/search/CategoryFilter.svelte
```diff
@@ -20,7 +20,12 @@ Implements active/inactive pill states matching Figma design
         "data-testid"?: string;
     }
 
-    let { selectedCategories = [], onCategoryChange, showLabel = true }: Props = $props();
+    let {
+        selectedCategories = [],
+        onCategoryChange,
+        showLabel = true,
+        "data-testid": dataTestid,
```
subiabre: `data-testid` attributes are dev-only remnants from a developer who used an external testing tool on his implementations. Please, simply remove them from the source code.

src/pages/[...locale]/index.astro
```diff
@@ -40,6 +38,7 @@ const { t, lang } = Astro.locals;
         <Stats />
         <CampaignsTabs />
         <Matchfunding />
+        <!-- Re-add the matching import when enabling either of these. -->
```
Comment (subiabre) [inline src/pages/[...locale]/index.astro]: Please, also leave as comments the import statements or at least the paths to the components as to avoid possible future ambiguity.

subiabre (changes requested):

PR #396, opened 2026-08-06
[feat] admin/comm#banners page + add refactor unified reusable Table component

Dayv1dDev:
---
  - https://app.asana.com/0/0/1217108196643803


src/i18n/locales/es.json
```diff
@@ -1612,6 +1552,54 @@
...
+      "comm": {
+        "title": "Comunicación",
+        "description": "Administra la comunicación de la plataforma",
+        "tabs": {
+          "banners": "Anuncios"
+        },
+        "pagination": {
+          "showing": "{{ items }} de {{ total }} resultados",
+          "noResults": "No se han encontrado resultados",
+          "unloaded": "Cargando..."
```
subiabre: Pagination is a very common product-domain feature, and there is nothing relevant to the /admin/comm or banners here. Why not move these keys to `common.pagination` or `domain.pagination` ? Please, also update the references keys in the other admin pages and remove the legacy keys from the es.json file.

src/components/admin/comm/banners/BannersForm.svelte
subiabre, line 1: This component implements a form, but it's not structured as such, which will cause problems to users relying on accessibility tools. Would also rename this component as `*Form` to follow the convention with other forms like `CreateProjectForm`.

Finally, I know the Figma design places the "save" button along the title, but this adds unexpected complexity with the current architecture of this page, makes the admin left sidebar top behave weirdly when the scroll position in near the button and feels out of place next to other forms in the application like the login, register or all the forms in a modal. Please, move the save button inside this same component and place it coherently to other forms.

src/components/admin/comm/banners/Banners.svelte
```diff
@@ -0,0 +1,38 @@
...
+        </Button>
+    {/if}
+</div>
+
+<div class="flex flex-col gap-6">
+    <BannersTabs currentTab={currentSubtab} onTabChange={handleTabChange} />
+    {#if currentSubtab === "fields"}
+        <BannersFields />
+    {:else if currentSubtab === "history"}
+        <BannersHistory rows={[]} /> // TODO: Replace with actual D1 data
```
subiabre: If the "TO-DO" note was meant as a comment it should have been written as `<!--TO-DO: Replace ...-->`

subiabre (changes requested):

src/components/admin/comm/banners/Banners.svelte
Dayv1dDev: Oops, I'm so used to JavaScript, my bad.

